PR #2, opened 2017-06-07
use dirname(@__FILE__) instead of Pkg.dir

tkelman:
this allows the package to be installed and loaded from elsewhere

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -9,3 +9,4 @@ IntervalTrees
 julia 0.5
 PairwiseListMatrices
 Twiddle
+DataStructures
--- a/test/io/abif.jl
+++ b/test/io/abif.jl
@@ -16,7 +16,7 @@
     end
 
     get_bio_fmt_specimens()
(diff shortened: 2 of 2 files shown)


Ward9250:
Can you make the PR use release/v0.5.0 as the base branch and then I can re-tag the release with this fix? It might as well go in the initial release!

Ward9250:
Thanks @tkelman this is great!

tkelman:
No problem, these are pretty minor. Guess one of your dependencies was bringing in DataStructures as a transitive dep.

How much of the code here is copied over from Bio.jl?

Ward9250:
Much of it is from Bio.jl, but stuff from Bio.jl that was slowly becoming depreciated has been given a push, and so some code is different. But the first release is to be as close to Bio.jl as possible as it is simply a step in the de-centralization of Bio.jl. I will continue some work which will probably differentiate it over the next few weeks.

PR #5, opened 2017-06-09
[WIP] Improve Julia 0.6 compatibility

bicycle1885:
Since Julia v0.6 is almost there we need to prepare BioSequences.jl for it.

### DIFF
--- a/src/bioseq/site_counting/site_counting.jl
+++ b/src/bioseq/site_counting/site_counting.jl
@@ -155,13 +155,14 @@ end
 
 function count_pairwise{P<:Position,A<:NucleicAcidAlphabets,N}(::Type{P}, seqs::Vararg{BioSequence{A},N})
     @assert N >= 2 "At least two sequences are required."
-    counts = Vector{bp_counter_type(P, A)}(Int((N * (N - 1)) / 2))
-    c = 1
-    @inbounds for i in 1:N, j in (i + 1):N
-        counts[c] = count(P, seqs[i], seqs[j])
-        c += 1
+    counts = Matrix{bp_counter_type(P, A)}(N, N)
+    for i in 1:N
+        counts[i,i] = zero(eltype(counts))
+        for j in (i+1):N
+            counts[i,j] = counts[j,i] = count(P, seqs[i], seqs[j])
+        end
     end
-    return PairwiseListMatrix(counts, false)
+    return counts
 end
 
 
@@ -196,7 +197,7 @@ end
 Count the number of sites of type `S`, between each possible pair of the sequences
 provided (`seqs`).
 
-This returns a PairwiseListMatrix object contining the output for each pairwise
+This returns a symmetric matrix contining the output for each pairwise
 operation.
 """
 function count_pairwise{S<:Site,N}(::Type{S}, seqs::Vararg{BioSequence,N})
--- a/src/bioseq/site_counting/site_counting.jl
+++ b/src/bioseq/site_counting/site_counting.jl
@@ -10,8 +10,8 @@
 # Site types
 # ----------
 
-abstract Site
-abstract Position <: Site
+@compat abstract type Site end
+@compat abstract type Position <: Site end
 
 """
 A `Certain` site describes a site where both of two aligned sites are not an
--- a/src/bioseq/site_counting/site_counting.jl
+++ b/src/bioseq/site_counting/site_counting.jl
@@ -68,13 +68,13 @@ for A in (DNAAlphabet, RNAAlphabet)
         # Certain
         @inline function bp_chunk_count(::Type{Certain}, ::Type{$A{4}}, x::UInt64)
             x = enumerate_nibbles(x)
-            x $= 0x1111111111111111
+            x = x ⊻ 0x1111111111111111
             return count_zero_nibbles(x)
         end
 
         @inline function bp_chunk_count(::Type{Certain}, ::Type{$A{4}}, a::UInt64, b::UInt64)
-            x = enumerate_nibbles(a) $ 0x1111111111111111
-            y = enumerate_nibbles(b) $ 0x1111111111111111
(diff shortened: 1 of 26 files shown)


bicycle1885:
I found PairwiseListMatrices.jl does not support Julia 0.6 and it is not trivial to do that. So, I'd like to drop the dependency on the package if possible. I think `PairwiseListMatrix` is used in a few places and it can be replaced with usual matrices.

Ward9250:
I have no issue with that, so long as my pairwise site counting stuff still works. So long as when PairwiseListMatrices is 0.6 compatible the code is changed back; storing something twice when it's not needed isn't optimal. Out of interest, what is the issue with PairwiseListMatrices and 0.6?

src/bioseq/site_counting/site_counting.jl
```diff
@@ -89,20 +89,20 @@ for A in (DNAAlphabet, RNAAlphabet)
 
         # Match
         @inline function bp_chunk_count(::Type{Match}, ::Type{$A{4}}, a::UInt64, b::UInt64)
-            return count_zero_nibbles(a $ b)
+            return count_zero_nibbles(a ⊻ b)
         end
 
         @inline function bp_chunk_count(::Type{Match}, ::Type{$A{2}}, a::UInt64, b::UInt64)
-            return count_zero_bitpairs(a $ b)
+            return count_zero_bitpairs(a ⊻ b)
```
Ward9250: Is this necessary? Is `$` being changed? It's easy to type the `$` character, not so easy to type \veebar. (The only reason I knew it was \veebar is because I know LaTeX! Others will not know).

bicycle1885:
There are many issues; macros, deprecated array constructions, broadcasting syntax, etc.

src/bioseq/site_counting/site_counting.jl
bicycle1885: `x $ y` is deprecated and now we should use `xor(x, y)` or `x ⊻ y`. You can use `\xor` to output `⊻` character on Julia 0.6.

Ward9250:
Ok, we'll use typical matrices for now. If it's just a case of time and manpower, if you want help with PairwiseListMatrices (I saw your fork deigozea's ;) ), let me know.

Ward9250:
Also not that if you would like these changes to go into the first BioSequences release v0.5 then use the release/v0.5.0 branch as the base for this PR and I'll re-tag v0.5.0, otherwise master is fine.

bicycle1885:
I think it's okay to keep the current release tag.

BTW, the branch name release/v0.5.0 looks weird to me. A branch is a tip of a sequence of commits and thus it may move after created. However, release/v0.5.0 looks like pointing to a commit and immovable. I think release-0.5 is a better name as we used in Bio.jl and other packages. Unlike branches, a tag points to a fixed commit and it is natural to name a tag like v0.5.0.

Ward9250:
I think release-X is a fine naming system we can use that to keep consistency. The release branches are not eternal anyway according to the [OneFlow](http://biojulia.github.io/Contributing/latest/help/branching_guide/) model and idiom that there should be one eternal branch: release branches are temporary and get merged back into master.

bicycle1885:
This is almost ready to merge. All deprecated messages are removed except one: "WARNING: deprecated syntax "_ as an rvalue around test.jl:913"." I couldn't locate the place it happened.

Eternal release branches are used in the development of Julia and it seems to work well. Also, it works well on other packages we maintain. So, I think there is no reason to change the development model.

Ward9250:
Release branches can be eternal so long as they are merged back into master as hot-fixes are made, cherry picking hot-fixes onto a release branch from master isn't as nice as applying the fix to the release branch, and then merging into master IMHO. This is because IIRC cherry-picking creates a new commit with a new hash, but the same content changes. Meaning history of release branches and master diverges, even when their content is made more similar.

bicycle1885:
> Cherry picking hot-fixes onto a release branch from master isn't as nice as applying the fix to the release branch, and then merging into master IMHO.

Why? Cherry-picking patches from master is very easy if they do not conflict with a release branch. If they conflict, the opposite (move commits from release to master) will also conflict and nothing gets easier.

I' m not perfectly sure what you mean. Could you explain the process we do when we need to release a bugfix version?

I think the most important advantage of having eternal release branches (e.g. release-0.5) is it is very clear to find which commits should be cherry-picked from the master branch. Since bug fix releases should not include new features, what we should do is just to pick bug fix commits from master and then cherry pick them onto the release branch. If we don't have any release branch, we need to carefully choose a branch point that does not include a new feature to tag a bugfix version.

Ward9250:
>Why? Cherry-picking patches from master is very easy if they do not conflict with a release branch. If they conflict, the opposite (move commits from release to master) will also conflict and nothing gets easier.

Conflicts aren't the issue, conflicts are conflicts, whatever method you use, they need resolving.

But AFAIK, cherry picking creates new commits with new hashes, with the same content. So two branches get the same change - master and release get the fix, but different commits and shas. This doesent scale well with many cherry picks - many fixes, and makes the history of master and a release more divergent than it needs to be.

>I' m not perfectly sure what you mean. Could you explain the process we do when we need to release a bug-fix version?

1). Realise a fix is needed:
2). Check out the release branch.
3). Make commits to apply fix.
4). Tag and release with a minor version increment.
5). Merge with master so master has the updates (note merging into master doesn't mean the release branch is removed).

bicycle1885:
Ah, it sounds good to me. I don' like merging because it complicates the commit history but it may be okay if merged branches are very short-lived (say <4 commits). Let's try it from the next release of BioSequences.jl.

bicycle1885:
I'll merge this if no objections.

bicycle1885:
Oops, I pushed "rebase and merge" instead of "squash and merge". Sorry for making the commit history noisy.

PR #6, opened 2017-06-09
use dirname(@__FILE__) instead of Pkg.dir in one more location

tkelman:
missed this one (not urgent, about to merge registration, this can be in the next tag)


bicycle1885:
Thank you!

PR #8, opened 2017-06-12
An eachkmer iterator fix

Ward9250:
@bicycle1885 This edits some lines of the eachkmer iterator, so as well as checking for ambigs, it also checks for gaps. Hope this is ok, if it is, the use of `each_kmer_impl` in some specific iterator I can write will actually allow me to do a lot of my work without the need for a whole new `Codon` type, which closes the issue and is a huge timesaving win!

### DIFF
--- a/src/eachkmer.jl
+++ b/src/eachkmer.jl
@@ -75,7 +75,7 @@ end
     k = kmersize(T)
     pos, kmer = state
     pos += it.step
-    has_ambiguous = false
+    ambig_or_gap = false
 
     # faster path: return the next overlapping kmer if possible
     if it.step < k && pos + k - 1 ≤ endof(it.seq)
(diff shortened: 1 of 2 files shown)


bicycle1885:
Good catch! I haven't noticed this problem

src/eachkmer.jl
```diff
@@ -75,27 +75,27 @@ end
...
+    ambig_or_gap = false
 
     # faster path: return the next overlapping kmer if possible
     if it.step < k && pos + k - 1 ≤ endof(it.seq)
         offset = k - it.step
         if it.step == 1
             nt = inbounds_getindex(it.seq, pos+offset)
             kmer = kmer << 2 | trailing_zeros(nt)
-            has_ambiguous |= isambiguous(nt)
+            ambig_or_gap |= (isambiguous(nt) || isgap(nt))
```
bicycle1885: Why not `iscertain(nt)`? A quick benchmark on my machine suggests no performance degradation by switching `isambiguous` with `iscertain`.

bicycle1885: Also, adding test cases to make sure gap skipping.

Ward9250: Good catch! I should have seen `iscertain` was applicable here, I used it instead of isambiguous and isgap in the other site counting stuff I've done. I'll make the iscertain change now, and add a test case in the morning - it's too late here!

Ward9250:
Ok here we go, a test explicitly checking a sequence with gaps. If this passes and is alright, I'd like to merge this fix when the checks are done.

bicycle1885:
👍  CI failure on Linux would not be a problem.

PR #12, opened 2017-06-23
Added diag_val methods

Ward9250:
Addresses the type-piracy comment by @tkelman by avoiding overloading Base.zero for tuples.


Ward9250:
Tests passing, I'll merge now in the interest of speed on account of GeneticVariation.jl

PR #13, opened 2017-07-06
use IterTools instead of Iterators

bicycle1885:
because Iterators.jl is deprecated: https://github.com/JuliaCollections/Iterators.jl.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,7 +6,7 @@ BufferedStreams
 Combinatorics
 Compat 0.18
 IndexableBitVectors
-Iterators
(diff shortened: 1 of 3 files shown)


bicycle1885:
@Ward9250 Did you tagged a new release? I guess so but I cannot find a pull request to METADATA. Perhaps attobot is not working.

Ward9250:
@bicycle1885 Yeah I've just removed and readded BioSequences to attobot and redid the release, now it's working :+1:

bicycle1885:
I see. Thank you very much 👍

PR #16, opened 2017-07-24
add position-weighted matrix search

bicycle1885:
This adds functionality to search motifs using position-weighted matrix (PWM), which is similar to `matchPWM` of [Biostrings](https://bioconductor.org/packages/release/bioc/html/Biostrings.html). 

TODO:
- [ ] ~support amino acid sequences~
- [x] docs
- [x] more tests
- [x] benchmark against Biostrings

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -2,7 +2,7 @@ using Documenter, BioSequences
 
 makedocs()
 deploydocs(
-    deps = Deps.pip("mkdocs", "pygments", "mkdocs-material"),
+    deps = Deps.pip("mkdocs", "pygments", "mkdocs-material", "python-markdown-math"),
     repo = "github.com/BioJulia/BioSequences.jl.git",
     julia = "0.6",
     osname = "linux",
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -23,7 +23,7 @@ extra_css:
   - assets/Documenter.css
 
 extra_javascript:
-  - https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML
+  - https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS_HTML
   - assets/mathjaxhelper.js
 
 # Extensions
@@ -33,6 +33,7 @@ markdown_extensions:
   - fenced_code
   - codehilite
   - def_list
+  - mdx_math
 
 # Page tree
 pages:
--- a/docs/src/search.md
+++ b/docs/src/search.md
@@ -215,6 +215,78 @@ Nullable{BioSequences.RE.RegexMatch{BioSequences.BioSequence{BioSequences.AminoA
 ```
 
 
+## Position weight matrix search
+
+A motif can also be specified using [position weight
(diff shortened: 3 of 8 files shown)


bicycle1885:
A quick benchmark suggests my implementation is significantly faster than that of Biostrings:

```
function searchall(seq, matrix)
    pwm = PWM{DNA}(matrix)
    score = maxscore(pwm) * 0.95
    ret = Int[]
    p = 0
    while (p = search(seq, pwm, score, p + 1)) != 0
        push!(ret, p)
    end
    return ret
end
```

BioSequences.jl
```
julia> length(chr1)
248956422

julia> size(matrix)
(4, 13)

julia> @time length(searchall(chr1, matrix))
  2.839495 seconds (18 allocations: 17.375 KiB)
725

```

Biostrings
```
> system.time(countPWM(pwm, chr1[[1]], min.score="95%"))
   user  system elapsed
  3.303   0.010   3.330
Warning message:
In .Call2("XString_match_PWM", pwm, subject, min.score, count.only,  :
  'subject' contains letters not in [ACGT] ==> assigned weight 0 to them
```

bicycle1885:
Okay, now this is ready to review.

jangevaare:
Could this be made even faster using StaticArrays for PWM and PFMs?

bicycle1885:
> Could this be made even faster using StaticArrays for PWM and PFMs?

I don't think so. I already use linear indexing to access an element in a matrix (https://github.com/BioJulia/BioSequences.jl/pull/16/files#diff-380589261798a9674fc004708668f4dcR293), which is the fastest way AFAIK.

> I wonder if this should instead be implemented in BioSymbols.jl for all abstract arrays?

This may be a sensible idea. But we need at least two kinds of indexing

bicycle1885:
BTW, indexing using a lookup table may be slower than using `trailing_zeros` because the lookup table will need cache (or memory) access while `trailing_zeros` will be compiled to [ffs](https://en.wikipedia.org/wiki/Find_first_set) in modern processors like this:
```
julia> @code_llvm trailing_zeros(DNA_C)

define i64 @julia_trailing_zeros_60822(i8) #0 !dbg !5 {
top:
  %1 = call i8 @llvm.cttz.i8(i8 %0, i1 false)
  %2 = zext i8 %1 to i64
  ret i64 %2
}

julia> @code_native trailing_zeros(DNA_C)
        .section        __TEXT,__text,regular,pure_instructions
Filename: nucleicacid.jl
        pushq   %rbp
        movq    %rsp, %rbp
Source line: 316
        movzbl  %dil, %eax
        orl     $256, %eax              ## imm = 0x100
        tzcntl  %eax, %eax
        popq    %rbp
        retq
        nopw    %cs:(%rax,%rax)

```

Ward9250:
To get the different indexing strategies, maybe we'd need a different wrapping type for an array. Using matrices from StaticArrays would be easy, as you could define a get and setindex for nucleotides where the array has 4 cols & 4 cols, as those are type parameters, and a get and setindex for such an array with 16 rows and 16 columns. To get similar behaviour for any other array would require a runtime check of the array dimensions.

Ward9250:
Is this ready to merge?

Ward9250:
I'll merge this, this morning (10:30 GMT), if there are no objections

PR #17, opened 2017-07-25
[Review] Drop Julia 0.5

Ward9250:


Ward9250:
Yeah I know I'm still WIP on it. 👍

bicycle1885:
I thought so, but just in case. I think updating the CI settings is the last thing before merging it.

BTW, I tagged BioSymbols.jl v1.2.0 including latest changes (https://github.com/JuliaLang/METADATA.jl/pull/10429).

Ward9250:
Awesome! 🎉

Ward9250:
Ok this is looking ready now. There are a couple of methods in the source which I wonder if they need to be parametric or not, as the type of sequence does not need to be constrained - the alphabet is left as 'A', and the 'A' then is not used in the method body, but we could always remove those later, the syntax is updated and works :+1:

Ward9250:
I'll merge and tag when the latest checks have gone through.

PR #19, opened 2017-08-02
Make minhash function for generic reader

kescobo:
Minor addition. Decided to use this with fastq files - works great. One thing: will only work if the reader has a `sequence()` defined for it. Is there any situation where this wouldn't work?

### DIFF
--- a/src/minhash.jl
+++ b/src/minhash.jl
@@ -93,17 +93,7 @@ function minhash(seqs::Vector{T}, k::Integer, s::Integer) where {T<:BioSequence}
     return MinHashSketch(kmerhashes, k)
 end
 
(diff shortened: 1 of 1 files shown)


bicycle1885:
`sequence` will work for FASTA, FASTQ, 2bit, SAM and BAM. It may fail if the sequence cannot be converted to a biological sequence but it is okay.

bicycle1885:
`continuous-integration/appveyor/branch — Waiting for status to be reported` is waiting for something forever. I'm not sure why.

Ward9250:
This looks good! And as @bicycle1885 says, if something fails in `sequence` as a result of getting the sequence for a FASTA record, it's probably supposed to (bad format and such).

PR #21, opened 2017-08-07
use Dict to store counts in Composition

bicycle1885:
This is a little bit slower than using Vector but it is able to store
counts of long k-mers with far less memory.

### DIFF
--- a/docs/src/search.md
+++ b/docs/src/search.md
@@ -221,22 +221,9 @@ Sequence composition can be easily calculated using the `composition` function:
 ```jldoctest
 julia> comp = composition(dna"ACGAG")
 DNA Composition:
-  DNA_Gap => 0
-  DNA_A   => 2
-  DNA_C   => 1
-  DNA_M   => 0
-  DNA_G   => 2
-  DNA_R   => 0
-  DNA_S   => 0
-  DNA_V   => 0
-  DNA_T   => 0
-  DNA_W   => 0
-  DNA_Y   => 0
-  DNA_H   => 0
-  DNA_K   => 0
-  DNA_D   => 0
-  DNA_B   => 0
-  DNA_N   => 0
+  DNA_A => 2
+  DNA_G => 2
+  DNA_C => 1
 
 julia> comp[DNA_A]
 2
--- a/src/composition.jl
+++ b/src/composition.jl
@@ -6,8 +6,14 @@
 # This file is a part of BioJulia.
 # License is MIT: https://github.com/BioJulia/BioSequences.jl/blob/master/LICENSE.md
 
+"""
+Sequence composition.
+
+This is a subtype of `Associative{T,Int}`, and the `getindex` method returns the
+number of occurrences of a symbol or a k-mer.
+"""
 struct Composition{T} <: Associative{T,Int}
-    counts::Vector{Int}
+    counts::Dict{T,Int}
 end
 
 function Composition(seq::BioSequence{A}) where {A<:Union{DNAAlphabet,RNAAlphabet}}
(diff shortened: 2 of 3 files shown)


Ward9250:
Could we have both versions? Unified by an API and abstract type, and then the best one for the use case applied?

bicycle1885:
We can, but I don't think it is a good idea since it makes the code more complicated. The new data structure I propose is still fast in most cases and the counting algorithm is not changed. So, the performance of `composition` does not change for `BioSequence`s:

master:
```
julia> @benchmark composition($seq)
BenchmarkTools.Trial:
  memory estimate:  224 bytes
  allocs estimate:  2
  --------------
  minimum time:     27.222 μs (0.00% GC)
  median time:      29.200 μs (0.00% GC)
  mean time:        30.333 μs (0.00% GC)
  maximum time:     212.823 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

```

this branch:
```
julia> @benchmark composition($seq)
BenchmarkTools.Trial:
  memory estimate:  800 bytes
  allocs estimate:  10
  --------------
  minimum time:     27.721 μs (0.00% GC)
  median time:      28.023 μs (0.00% GC)
  mean time:        30.111 μs (0.00% GC)
  maximum time:     137.682 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

```

The difference gets large for `Kmer`s but using `composition` for a `Kmer` value is not best in terms of performance because it does some allocation. Using `count_{a,c,g,t}` is the best choice in that case.

master:
```
julia> @benchmark composition(kmer"ACGTAGTA")
BenchmarkTools.Trial:
  memory estimate:  224 bytes
  allocs estimate:  2
  --------------
  minimum time:     42.198 ns (0.00% GC)
  median time:      46.022 ns (0.00% GC)
  mean time:        54.982 ns (11.49% GC)
  maximum time:     1.431 μs (86.91% GC)
  --------------
  samples:          10000
  evals/sample:     987

julia> @benchmark BioSequences.count_a(kmer"ACGTAGTA")
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.896 ns (0.00% GC)
  median time:      2.050 ns (0.00% GC)
  mean time:        2.210 ns (0.00% GC)
  maximum time:     35.214 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

```

this branch:
```
julia> @benchmark composition(kmer"ACGTAGTA")
BenchmarkTools.Trial:
  memory estimate:  592 bytes
  allocs estimate:  9
  --------------
  minimum time:     255.330 ns (0.00% GC)
  median time:      271.447 ns (0.00% GC)
  mean time:        322.011 ns (9.04% GC)
  maximum time:     6.661 μs (85.65% GC)
  --------------
  samples:          10000
  evals/sample:     352

```

Ward9250:
That sounds fair enough. Perhaps I should have a look at using Composition with the fast pairwise counting mechanism I built.

Ward9250:
If I dict is used for the composition, is there any point to the wrapping associative type anymore? It original purpose seems to be to act like a dict, but using a vector underneath, now a dict is used underneath, is there a point to the wrapping type anymore?

bicycle1885:
Yeah, that makes sense, but the semantic is significantly different from dicts. For example, `getindex` is overloaded to return zero for non-exist elements, which is similar to `DefautDict` of DataStructures.jl. Also, `merge` adds two counts instead of overwriting the former with the latter. `Composition` is convertible to `Dict`, too.

Ward9250:
I'm convinced! 😆  Look forward to the merge!!

PR #22, opened 2017-08-08
[WIP] Add unique sequence counting function

Ward9250:
This adds a method which computes counts of the unique sequences in a set of sequences:

```julia
```jlcon
julia> a = dna"AAAAAAAATTTTTT"
14nt DNA Sequence:
AAAAAAAATTTTTT

julia> b = dna"AAAAAAAATTTTTT"
14nt DNA Sequence:
AAAAAAAATTTTTT

julia> c = a[5:10]
6nt DNA Sequence:
AAAATT

julia> count_unique_sequences(a, b, c)
Dict{BioSequences.BioSequence{BioSequences.DNAAlphabet{4}},Int64} with 2 entries:
  AAAATT         => 1
  AAAAAAAATTTTTT => 2
```


Ward9250:
@bicycle1885 I'm wondering if this should be renamed to composition or something similar that can be treated as a composition. The reason I wrote this is, if you have a set of sequences from a population, often you want to know the unique sequences (i.e the alleles/genotypes) and count how many of each there are - allowing you to compute allele frequencies from the population's sequence data. In this way, it's analogous to a Composition method counting the unique kmers in a sequence, but rather it is counting the unique sequences in a population sample.

bicycle1885:
The most generic approach would be:
```julia
function composition(iter)
    counts = Dict{eltype(iter),Int}()
    for item in iter
        counts[item] += get(counts, item, 0) + 1
    end
    return Composition(counts)
end
```

This is based on top of #22.

Ward9250:
Thanks, glad to see you think this is a useful addition to `composition` :+1:, I'll finish it up today.

Ward9250:
One of the appveyor branches seems to be taking ages to load, but all others have passed. Will merge.

Issue #27, opened 2017-09-28
Further generalise the bitparallel counting code


BenJWard:
There are other operations I've seen need of - computing the number of segregating sites in a population sample of sequences and so on, that is not purely counting the number of a kind of site between two sequences, but that CAN be done, with a bit of re-working, with the bit-parallel counting methods already developed here. So previously I generalised the code, to counting different kinds of sites bit-parallel, now I'm going to generalise it further to let it do much more than just pairwise site counting. Related PR to come.

timbitz:
@Ward9250, @bicycle1885, I needed to calculate gc_content quickly for fastq reads and was looking into the current implementation and saw it was iterating through each nucleotide to calculate gc content or test for ambiguity (which means lots of needless masking and shifting?).  I figured that some bitparallel code would be faster, so I wrote some: https://gist.github.com/timbitz/cb728a9dcddf49f3c3a29e81102f752f
The implementations in that Gist are 10-30x faster than iterating through every nucleotide, for long sequences.

```julia
julia> longseq = dna"ATGACAAGATGACAGATAGACAGATAAGACAAAAATGGGT" ^ 1_000_000
40000000nt DNA Sequence:
ATGACAAGATGACAGATAGACAGATAAGACAAAAATGGG…TGACAAGATGACAGATAGACAGATAAGACAAAAATGGGT

julia> longseq_ambig = deepcopy(longseq)
40000000nt DNA Sequence:
ATGACAAGATGACAGATAGACAGATAAGACAAAAATGGG…TGACAAGATGACAGATAGACAGATAAGACAAAAATGGGT

julia> longseq_ambig[end] = DNA_N
DNA_N

julia> @time parallel_hasambiguity(longseq)
  0.008842 seconds (4 allocations: 160 bytes)
false

julia> @time parallel_hasambiguity(longseq_ambig)
  0.008782 seconds (4 allocations: 160 bytes)
true

julia> @time hasambiguity(longseq)
  0.113598 seconds (4 allocations: 160 bytes)
false

julia> @time hasambiguity(longseq_ambig)
  0.110090 seconds (4 allocations: 160 bytes)
true

julia> @time parallel_gc_content(longseq)
  0.004346 seconds (5 allocations: 176 bytes)
0.35

julia> @time gc_content(longseq)
  0.132628 seconds (5 allocations: 176 bytes)
0.35
```

[...]

PR #29, opened 2017-11-08
implement bit-parallel GC counter

bicycle1885:
~Still an experimental implementation.~

This is a faster version of `gc_content` using a bit-parallel algorithm as I mentioned in https://github.com/BioJulia/BioSequences.jl/issues/27#issuecomment-342789976. I was inspired to try a bit-parallel algorithm by @timbitz, so I think you are interested in this pull request.

```
julia> srand(1234); seq = randdnaseq(1_000_000);

julia> @benchmark count(isGC, seq)
BenchmarkTools.Trial:
  memory estimate:  16 bytes
  allocs estimate:  1
  --------------
  minimum time:     7.788 ms (0.00% GC)
  median time:      8.220 ms (0.00% GC)
  mean time:        8.548 ms (0.00% GC)
  maximum time:     12.825 ms (0.00% GC)
  --------------
  samples:          585
  evals/sample:     1

julia> @benchmark BioSequences.count_gc(seq)
BenchmarkTools.Trial:
  memory estimate:  16 bytes
  allocs estimate:  1
  --------------
  minimum time:     171.186 μs (0.00% GC)
  median time:      171.251 μs (0.00% GC)
  mean time:        181.268 μs (0.00% GC)
  maximum time:     600.158 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

```

### DIFF
--- a/src/sequence.jl
+++ b/src/sequence.jl
@@ -118,7 +118,7 @@ Calculate GC content of `seq`.
 """
 function gc_content(seq::Sequence)
     if !(eltype(seq) <: NucleicAcid)
-        error("elements must be nucleotides")
+        throw(ArgumentError("not a nucleic acid sequence"))
     end
     if isempty(seq)
         return 0.0
--- a/src/sequence.jl
+++ b/src/sequence.jl
@@ -107,6 +107,10 @@ end
 Base.findfirst(seq::Sequence, val) = findnext(seq, val, 1)
 Base.findlast(seq::Sequence, val) = findprev(seq, val, endof(seq))
 
+
+# GC content
+# ----------
+
 """
     gc_content(seq::Sequence)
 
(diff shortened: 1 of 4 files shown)


bicycle1885:
I've added a two-bit version and some tests. I think this is in a good shape to review.

timbitz:
Awesome @bicycle1885!  Sounds good, I'll take a look tomorrow morning.

src/sequence.jl
```diff
@@ -107,30 +107,30 @@ end
...
-
     if isempty(seq)
         return 0.0
     else
-        return gc / length(seq)
+        return count_gc(seq) / length(seq)
     end
 end
 
+function count_gc(seq::Sequence)
```
timbitz: Should this have the same name as count_gc in operators.jl?

timbitz: Ohh i c. Its just for arbitrary other nucleic acid alphabets that don't fall into DNAAlphabet or RNAAlphabet... Cool.

bicycle1885: That's right. Thank you for reviewing it.

timbitz: Happy to review anytime!

Issue #23, opened 2017-08-09
Add split function


onetonfoot:
I was trying to add the split function for BioSequences. 

```julia

Base.split(str::T, splitter; limit::Integer=0, keep::Bool=true) where {T<:Sequence} =
    Base._split(str, splitter, limit, keep, T[])
function Base._split(str::Sequence,splitter, limit::Integer , keep_empty::Bool ,strs::Array)

    i = start(str)
    n = endof(str)
    r = search(str,splitter,i)
    j,k = first(r), last(r) + 1

    while 0 < j <= n && length(str) != limit-1
        if i < k
            if keep_empty || i < j
                push!(strs,str[i:j-1])
            end
            i = k
        end
        (k <= j) && (k = j + 1)
        r = search(str,splitter,k)
        j, k = first(r), last(r) + 1
    end

    if keep_empty || !done(str,i)
        push!(strs, str[i:end])
    end

    return strs
end
```

It was based on the one in base shown below.  

```julia

split(str::T, splitter; limit::Integer=0, keep::Bool=true) where {T<:AbstractString} =
    _split(str, splitter, limit, keep, SubString{T}[])
function _split(str::AbstractString, splitter, limit::Integer, keep_empty::Bool, strs::Array)
    i = start(str)
    n = endof(str)
    r = search(str,splitter,i)
    j, k = first(r), nextind(str,last(r))
    while 0 < j <= n && length(strs) != limit-1
        if i < k
            if keep_empty || i < j
                push!(strs, SubString(str,i,prevind(str,j)))
            end
            i = k
        end
        (k <= j) && (k = nextind(str,j))
        r = search(str,splitter,k)
        j, k = first(r), nextind(str,last(r))
    end
    if keep_empty || !done(str,i)
        push!(strs, SubString(str,i))
    end
    return strs
end
```

It seems to work fine with DNASequences and RNASequences.

```julia
split(dna"AAAAACACAAAACATATATATC",dna"CA",keep=true)

4-element Array{BioSequences.BioSequence{BioSequences.DNAAlphabet{4}},1}:
 AAAAA   
         
 AAA     
 TATATATC
```

```julia
split(dna"AAAAACACAAAACATATATATC",dna"CA",keep=false)

3-element Array{BioSequences.BioSequence{BioSequences.DNAAlphabet{4}},1}:
 AAAAA   
 AAA     
 TATATATC
```

[...]

PR #31, opened 2017-11-18
Message printed when sequence is empty.

Ward9250:
In response to comments on issue #23.

### DIFF
--- a/src/sequence.jl
+++ b/src/sequence.jl
@@ -219,7 +219,7 @@ ungap!(seq::Sequence) = filter!(x -> x != gap(eltype(seq)), seq)
 # Printers
 # --------
 
-function Base.print(io::IO, seq::Sequence; width::Integer=0)
+function Base.print(io::IO, seq::Sequence; width::Integer = 0)
     col = 1
     for x in seq
         if width > 0 && col > width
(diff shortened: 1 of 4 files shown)


src/sequence.jl
```diff
@@ -241,19 +241,23 @@ end
...
-        half = div(width, 2)
-        for i in 1:half-1
-            print(io, seq[i])
-        end
-        print(io, '…')
-        for i in endof(seq)-half+2:endof(seq)
-            print(io, seq[i])
-        end
+    if length(seq) == 0
+        print(io, "< EMPTY SEQUENCE >")
```
bicycle1885: Adding at least one test case would be good.

Ward9250: 👍 I will do this soon

test/refseq/print.jl
```diff
@@ -3,11 +3,18 @@
...
 
-    buf = IOBuffer()
     print(buf, ReferenceSequence(dna"A"^100))
     @test String(take!(buf)) == "A"^100
 end
+
+@testset "Show" begin
+    buf = IOBuffer()
+    show(buf, ReferenceSequence(dna""))
+    @test String(take!(buf)) == "0nt Reference Sequence:\n"
```
bicycle1885: This will show "< EMPTY SEQUENCE >" as well.

PR #33, opened 2018-02-19
Bugfix: Writing FASTA sequences when width <= 0

Ward9250:
A typo / bracket error prevented writing sequences to file when the width of the FASTA was <= 0. This fixes that error.


Ward9250:
Tests pass for this.
Since this is a bug-fix and somewhat critical for anyone wanting to write FASTA's I'll merge now and do a bugfix release.

PR #41, opened 2018-08-13
Julia 1.0 Compatibility

dcjones:
This was an ordeal, and ended up being a huge patch, but should bring sequences up to date with julia 1.0 changes.

Needs a couple other PRs to work:
  * https://github.com/BioJulia/BioSymbols.jl/pull/30
  * https://github.com/BioJulia/BioCore.jl/pull/18

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,7 +4,7 @@ BioCore 2.0.0
 BioSymbols 3.0.1
 BufferedStreams
 Combinatorics
-IndexableBitVectors
+IndexableBitVectors 1.0.0
 IntervalTrees
 Twiddle 1.0.0
 DataStructures
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,5 +6,5 @@ BufferedStreams
 Combinatorics
 IndexableBitVectors
 IntervalTrees
-Twiddle
+Twiddle 2.0.0
 DataStructures
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.7
-Automa
+Automa 0.7
 BioCore 2.0.0
 BioSymbols 3.0.1
 BufferedStreams
@@ -8,4 +8,3 @@ IndexableBitVectors
 IntervalTrees
 Twiddle
 DataStructures
-Nullables
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.7
 Automa
-BioCore
-BioSymbols 2.0
+BioCore 2.0.0
+BioSymbols 3.0.1
 BufferedStreams
 Combinatorics
 IndexableBitVectors
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,7 +5,7 @@ BioSymbols 2.0
 BufferedStreams
 Combinatorics
 IndexableBitVectors
-IterTools 0.1
 IntervalTrees
 Twiddle
 DataStructures
+Nullables
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.7
 Automa 0.7
 BioCore 2.0.2
-BioSymbols 3.0.1
+BioSymbols 3.1.0
 BufferedStreams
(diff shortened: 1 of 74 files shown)


BenJWard:
Thanks for taking the time to do this @dcjones! I'll merge when the other two PRs are in :+1:

REQUIRE
```diff
@@ -1,11 +1,11 @@
...
+BioCore 2.0.0
+BioSymbols 3.0.1
 BufferedStreams
 Combinatorics
 IndexableBitVectors
-IterTools 0.1
 IntervalTrees
 Twiddle
 DataStructures
+Nullables
```
bicycle1885: It is okay to have this now but I think that in general we should not depend on deprecated/removed features. We may need a second round to review our APIs.

BenJWard:
@dcjones To take this furthur and get CI passes I'm currently waiting on YAML, the version 1.0 changes PR is waiting on a Codecs release, which is here: https://github.com/JuliaLang/METADATA.jl/pull/16489, but as it's not on attobot it's not moving very far. Can you head to the PR and usher them along a bit, or turn on attobot for Codecs and YAML? :)

BenJWard:
@dcjones Ok I've made a series of commits trying to get tests to pass. I've got them passing for 1.0, but curously not for 0.7. What's making me tear my hair out is julia complains "ERROR: LoadError: LoadError: LoadError: syntax: invalid variable expression in "where"". Whilst pointing to "in expression starting at /home/travis/build/BioJulia/BioSequences.jl/src/bioseq/constructors.jl:65". At which is a constructor, which does not make use of the "where" expression anywhere in it's definition. What do you think?

jgreener64:
So I had a go and could reproduce your issue @BenJWard. A MWE without imports that works on v0.7 but not on v1.0 is:

```julia
abstract type Alphabet end
abstract type Sequence end

struct DNAAlphabet{n} <: Alphabet end

mutable struct BioSequence{A<:Alphabet} <: Sequence
    data::Vector{UInt64}  # encoded character sequence data
    part::UnitRange{Int}  # interval within `data` defining the (sub)sequence
    shared::Bool          # true if and only if `data` is shared between sequences

    function BioSequence{A}(data::Vector{UInt64},
                            part::UnitRange{Int},
                            shared::Bool) where A
        return new(data, part, shared)
    end
end

function BioSequence{DNAAlphabet{4}}(seq::BioSequence{DNAAlphabet{2}}) # errors here
    newseq = BioSequence{vim}(length(seq))
    #for (i, x) in enumerate(seq)
    #    unsafe_setindex!(newseq, x, i)
    #end
    return newseq
end
```

It seems that if you replace the offending line with

```julia
function BioSequence{T}(seq::BioSequence{DNAAlphabet{2}}) where {T<:DNAAlphabet{4}}
```

then it seems to work, which is strange. I made a PR to implement this across this file and tests seem to pass on both v0.7 and v1.0.

Feel free to use it if you like, but I'm wondering if this is a Julia issue?

BenJWard:
Thanks @jgreener64! You're option made me think of another alternative, which also passes both of 0.7 and 1.0!

```julia
abstract type Alphabet end
abstract type Sequence end

struct DNAAlphabet{n} <: Alphabet end

mutable struct BioSequence{A<:Alphabet} <: Sequence
    data::Vector{UInt64}  # encoded character sequence data
    part::UnitRange{Int}  # interval within `data` defining the (sub)sequence
    shared::Bool          # true if and only if `data` is shared between sequences

    function BioSequence{A}(data::Vector{UInt64},
                            part::UnitRange{Int},
                            shared::Bool) where A
        return new(data, part, shared)
    end
end

function (::Type{BioSequence{DNAAlphabet{4}}})(seq::BioSequence{DNAAlphabet{2}})
    newseq = BioSequence{DNAAlphabet{4}}(length(seq))
    for (i, x) in enumerate(seq)
        unsafe_setindex!(newseq, x, i)
    end
    return newseq
end
```

BenJWard:
Merging into the `version-1.0` branch now. Release to follow shortly.

PR #47, opened 2018-10-10
Allow translate(seq) to take a DNASequence

kdyrhage:
# Allow translate(seq) to take a DNASequence
Currently `translate(seq::DNASequence)` throws a MethodError. While `translate(convert(RNASequence, seq))` might make sense biologically, it is unnecessarily verbose. I propose doing this conversion automatically.

## Types of changes

* [x] :sparkles: New feature (A non-breaking change which adds functionality).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- Automatic convertion of `DNASequence` to `RNASequence` when translating sequences.
+- Automatic conversion of `DNASequence` to `RNASequence` when translating sequences.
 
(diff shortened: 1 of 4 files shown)


kdyrhage:
Sounds like a good idea. There is no method for `translate(::BioSequence{RNAAlphabet{2}})` either, though, so I'm adding that as well.

Something else I was thinking of is adding a keyword argument for changing alternative start codons to `AA_M`. `ncbi_trans_table` doesn't contain this information, so the easiest thing would be to just always change the first amino acid if the flag is set. Something like `convert_start_codon::Bool=false`. How does that sound?

BenJWard:
Why do you want to convert the first codon to AA_M?

kdyrhage:
Because alternative start codons in bacteria are still recognised by fMet-tRNA, not by whatever tRNA they are usually recognised by.

BenJWard:
@kdyrhage All the boxes are ticked, and CI is running. When it passes I'm happy to merge this, unless there is anything else furthur that you want to do.

Issue #46, opened 2018-10-08
Faster computation of Kmer composition


jakobnissen:
Kmer composition is relatively widely used. The reason for using kmers rather than arbitrary length sequences are almost often the superior speed of kmers, hence it's an area worth optimizing.

Currently in BioSequences, calculating the kmer composition of a sequence relies on iterate(x::EachKmerIterator). However, rather than relying on a kmer iterator, counting kmers by a direct loop over each nucleotide of the sequence is significantly faster for longer sequences. In a test implementation I made, the kmer composition can be counted ~4x faster for long sequences (10k nt) using a direct loop, compared to the current BioSequences implementation.

I suggest changing the implementation of the method
`composition(iter::EachKmerIterator{T}) where {T<:Kmer}`
Such that it's calculated by direct iteration.

#### Advantages
- Quicker computation

#### Disadvantages
- Does not rely on the existing implementation of BioSequences.EachKmerIterator. This means that if the underlying structure of sequences change, updating BioSequences.EachKmerIterator will not be sufficient to making these methods work.

I have created an implementation of my proposed change at:
https://github.com/jakobnissen/biobag/blob/master/kmercount_newimpl.jl

__Note!__: I have not extensively tested my implementation for all the corner cases - I can write the test code if this proposal is accepted.

BenJWard:
Hey @jakobnissen, Have you identified why the iterators are not faster than the direct loop? Is it something that can be fixed about the iterators or do you think there's a general issue with iterators being slower than direct loops?

jakobnissen:
I don't know exactly what makes the iterator slower. Some weeks ago, I implemented my own biological symbols, biological sequences and Kmer library from scratch. Even though my implementation of both the biological sequence and my KmerIterator object differed from that in BioSymbols and BioSequences, I saw the same ~3x performance difference.
On the other hand the direct loop also uses `for i in 1:n`, which itself implements `iterate`, so I'm not sure it's something intrinsic with the iteration protocol.

BenJWard:
This is surprising as the iterate method on BioSequences tries to do some things which are reasonably smart:

```
function Base.iterate(
        it::EachKmerIterator{T},
        state::Tuple{Int, UInt64}) where {T}

    k = kmersize(T)
    pos, kmer = state
    isok = true

    # faster path: return the next overlapping kmer if possible
    if it.step < k && pos + k - 1 ≤ lastindex(it.seq) && pos != it.start
        offset = k - it.step
[...]

PR #49, opened 2018-10-12
Type specialization for faster kmer iteration

jakobnissen:
# Type specialization for faster kmer iteration and misc changes

## Types of changes

* [:heavy_check_mark: ] :sparkles: New feature (A non-breaking change which adds functionality).
* [ :heavy_check_mark:] :bug: Bug fix (A non-breaking change, which fixes an issue).
* [ :heavy_check_mark:] :boom: Breaking change (fix or feature that would cause existing functionality to change).

This pull requests has a variety of changes:

### Improved Kmer iteration speed

As discussed in [Issue 46](https://github.com/BioJulia/BioSequences.jl/issues/46), the previous implementation of `iterate` on `EachKmerIterator` was general, and did not specialize for the special cases of two-bit nucletide sequences, nor for cases where the iterator's `step` was 1.

Here, I have added three new types:
`AbstractKmerIterator`
`EveryKmerIterator <: AbstractKmerIterator`, for step == 1
`SpacedKmerIterator <: AbstractKmerIterator`, for step != 1

and removed the old `EachKmerIterator` type. Further, I have added a `length` method for 2-bit nucleotide kmer iterators, since their length can be known beforehand.

**Note**: It might be useful for the `each` function to take start and stop keywords, I have not added them in order to not change the interface.

**Justification**
Many bioinformatics workflows rely on kmers, precisely because they are fast. By changing the iteration protocol to improve speeds, many current and potential downstream applications of kmers such as kmer composition analysis or kmer distribution will be significantly sped up.

**Benchmark**
The performance difference depends on several factors: Sequence type, value of K, value of `step` and sequence length. 

In general, the speed relative to the current implementation is 4.5x in the typical case, 1.2x in the worst case and 27x in the best case.

In order to benchmark, I've used this simple function:
```
function bm(x)
    n = 0
    for i in x
        n += 1
    end
    return n
end
```

With the following results:
Before:
```
Len: 100    Bit: 2  K: 4    Step: 1    772.761 ns (0 allocations: 0 bytes)
Len: 100    Bit: 2  K: 4    Step: 4    323.026 ns (0 allocations: 0 bytes)
Len: 100    Bit: 2  K: 30   Step: 1    639.707 ns (0 allocations: 0 bytes)
Len: 100    Bit: 2  K: 30   Step: 4    317.760 ns (0 allocations: 0 bytes)
Len: 100    Bit: 4  K: 4    Step: 1    659.629 ns (0 allocations: 0 bytes)
Len: 100    Bit: 4  K: 4    Step: 4    280.278 ns (0 allocations: 0 bytes)
Len: 100    Bit: 4  K: 30   Step: 1    607.361 ns (0 allocations: 0 bytes)
Len: 100    Bit: 4  K: 30   Step: 4    276.466 ns (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 4    Step: 1    776.198 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 4    Step: 4    313.785 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 4    Step: 100   12.514 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 30   Step: 1    781.717 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 30   Step: 4    349.243 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 30   Step: 100   62.372 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 4    Step: 1    659.685 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 4    Step: 4    266.367 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 4    Step: 100   10.692 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 30   Step: 1    759.375 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 30   Step: 4    310.910 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 30   Step: 100   50.230 μs (0 allocations: 0 bytes)
```
After
```
Len: 100    Bit: 2  K: 4    Step: 1    57.360 ns (0 allocations: 0 bytes)
Len: 100    Bit: 2  K: 4    Step: 4    62.466 ns (0 allocations: 0 bytes)
Len: 100    Bit: 2  K: 30   Step: 1    55.463 ns (0 allocations: 0 bytes)
Len: 100    Bit: 2  K: 30   Step: 4    70.036 ns (0 allocations: 0 bytes)
Len: 100    Bit: 4  K: 4    Step: 1   151.853 ns (0 allocations: 0 bytes)
Len: 100    Bit: 4  K: 4    Step: 4   131.818 ns (0 allocations: 0 bytes)
Len: 100    Bit: 4  K: 30   Step: 1   148.653 ns (0 allocations: 0 bytes)
Len: 100    Bit: 4  K: 30   Step: 4   123.503 ns (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 4    Step: 1    28.718 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 4    Step: 4    35.829 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 4    Step: 100   1.460 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 30   Step: 1    28.694 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 30   Step: 4    42.997 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 2  K: 30   Step: 100  18.374 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 4    Step: 1   143.237 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 4    Step: 4   121.794 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 4    Step: 100   4.979 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 30   Step: 1   143.241 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 30   Step: 4   121.809 μs (0 allocations: 0 bytes)
Len: 100000 Bit: 4  K: 30   Step: 100  35.020 μs (0 allocations: 0 bytes)
```

### Disallow instantiation of zero-mers and negative kmers
__This change is breaking__

Previously, kmers with K of 0 or negative numbers could be instantiated. With these changes, attempting to do so throws an ArgumentError.
The rationale is both that zero or negative kmers make no sense biologically and thus should not be used, and also that they present as edge cases which complicates the code and leads to unexpected errors.

### Remove unused dependency DataStructures
Previously, DataStructures was listed as a dependency and imported in `minhash.jl` but not actually used in any code. I have simply removed DataStructures from `REQUIRE` and `minhash.jl`. This has no effect on the code.

### Change documentation and testing for reverse complementation
On the existing master branch, the documentation for `reverse_complements` states it leaves ambiguous nucleotides as-is, while in actuality, the existing code handles ambiguous nucleotides correctly. Also, the correct complementation and reverse-complementation of ambiguous nucleotides other than N are not tested for. I have amended the documentation and added test cases for ambiguous nucleotides.

### Minor speed increase in Composition for K > 8
Kmer Composition now increments its Dict more effectively resulting in a small speed increase. Other than the modest increase in speed, this is an inconsequential single-line change.

- [:heavy_check_mark: ] :art: The changes implemented is consistent with the [julia style guide](https://docs.julialang.org/en/stable/manual/style-guide/).
- [ :heavy_check_mark:] :blue_book: I have updated and added relevant docstrings, in a manner consistent with the [documentation styleguide](https://docs.julialang.org/en/stable/manual/documentation/).
- [:heavy_check_mark: ] :blue_book: I have added or updated relevant user and developer manuals/documentation in `docs/src/`.
- [:heavy_check_mark: ] :ok: There are unit tests that cover the code changes I have made.
- [:heavy_check_mark: ] :ok: The unit tests cover my code changes AND they pass.
- [:heavy_check_mark: ] :pencil: I have added an entry to the `[UNRELEASED]` section of the manually curated `CHANGELOG.md` file for this repository.
- [:heavy_check_mark: ] :ok: All changes should be compatible with the latest stable version of Julia.
- [:heavy_check_mark: ] :thought_balloon: I have commented liberally for any complex pieces of internal code.

### DIFF
--- a/src/eachkmer.jl
+++ b/src/eachkmer.jl
@@ -81,9 +81,18 @@ function each(::Type{Kmer{T,K}}, seq::Sequence, step::Integer=1) where {T,K}
     end
 end
 
-eachkmer(seq::BioSequence{A}, K::Integer, step::Integer=1) where {A<:DNAAlphabet} = each(DNAKmer{Int(K)}, seq, step)
-eachkmer(seq::BioSequence{A}, K::Integer, step::Integer=1) where {A<:RNAAlphabet} = each(RNAKmer{Int(K)}, seq, step)
-eachkmer(seq::ReferenceSequence, K::Integer, step::Integer=1) = each(DNAKmer{Int(K)}, seq, step)
+function eachkmer(seq::BioSequence{A}, K::Integer, step::Integer=1) where {A<:DNAAlphabet}
+    Base.depwarn("eachkmer is depreceated: type instability means it is too slow. Please use each(::Type{Kmer{T,K}}, seq, step) instead", :eachkmer)
+    return each(DNAKmer{Int(K)}, seq, step)
+end
+function eachkmer(seq::BioSequence{A}, K::Integer, step::Integer=1) where {A<:RNAAlphabet}
+    Base.depwarn("eachkmer is depreceated: type instability means it is too slow. Please use each(::Type{Kmer{T,K}}, seq, step) instead", :eachkmer)
+    return each(RNAKmer{Int(K)}, seq, step)
+end
+function eachkmer(seq::ReferenceSequence, K::Integer, step::Integer=1)
+    Base.depwarn("eachkmer is depreceated: type instability means it is too slow. Please use each(::Type{Kmer{T,K}}, seq, step) instead", :eachkmer)
+    return each(DNAKmer{Int(K)}, seq, step)
+end
 
 Base.eltype(::Type{<:AbstractKmerIterator{T,S}}) where {T,S} = Tuple{Int,T}
 Base.IteratorSize(::Type{<:AbstractKmerIterator{T,S}}
--- a/src/eachkmer.jl
+++ b/src/eachkmer.jl
@@ -17,7 +17,9 @@
 # increment is how far the index jumps ahead when going to the next kmer.
 # for close kmers where no jump is possible, it's 1, else it can be arbitrary high.
 #
-#       |---------- step --------|
+# For a kmeriterator that first emits xxxxxxx, then zzzzzzz:
+#
+#       |--------- step ---------|
 #       xxxxxxx                  zzzzzzz
 # -------------------------------------------------------------
 #           yyyyyyy
@@ -65,7 +67,7 @@ end
 function each(::Type{Kmer{T,K}}, seq::Sequence, step::Integer=1) where {T,K}
     if eltype(seq) ∉ (DNA, RNA)
         throw(ArgumentError("element type must be either DNA or RNA nucleotide"))
-    elseif !(0 ≤ K ≤ 32)
+    elseif !(1 ≤ K ≤ 32)
         throw(ArgumentError("k-mer length must be between 0 and 32"))
     elseif step < 1
         throw(ArgumentError("step size must be positive"))
@@ -84,9 +86,15 @@ eachkmer(seq::BioSequence{A}, K::Integer, step::Integer=1) where {A<:RNAAlphabet
 eachkmer(seq::ReferenceSequence, K::Integer, step::Integer=1) = each(DNAKmer{Int(K)}, seq, step)
 
 Base.eltype(::Type{<:AbstractKmerIterator{T,S}}) where {T,S} = Tuple{Int,T}
-Base.IteratorSize(::Type{<:AbstractKmerIterator{T,S}}) where {T,S} = Base.SizeUnknown()
+Base.IteratorSize(::Type{<:AbstractKmerIterator{T,S}}
+) where {T,S<:Union{ReferenceSequence, BioSequence{<:FourBitNucs}}} = Base.SizeUnknown()
+Base.IteratorSize(::Type{<:AbstractKmerIterator{T,S}}
+) where {T,S<:BioSequence{<:TwoBitNucs}} = Base.HasLength()
 
-positions(::Type{Kmer{T, K}}) where {T, K} = max(1, K) # Only to handle K = 0
+function Base.length(it::AbstractKmerIterator{T,S}) where {T,S<:BioSequence{<:TwoBitNucs}}
+    return max(0, fld(it.stop - it.start + 1 - kmersize(T), step(it)) + 1)
+end
+
 Base.step(x::EveryKmerIterator) = 1
 Base.step(x::SpacedKmerIterator) = x.step
 
@@ -106,7 +114,7 @@ function Base.iterate(it::AbstractKmerIterator{T,S}) where {T,S<:BioSequence{<:T
         @inbounds val = kmerbits[nt + 1]
         kmer = kmer << 2 | val
         filled += 1
-        if filled == positions(T)
+        if filled == kmersize(T)
             return (1, T(kmer)), (i, kmer)
         end
         i += 1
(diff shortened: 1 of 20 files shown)


BenJWard:
This is looking very nice, somethings up with a few tests but this is exciting!

BenJWard:
Are you finished with this @jakobnissen, or do you still have a bit more to do?

src/eachkmer.jl
```diff
@@ -35,71 +67,140 @@ end
...
-        nt = inbounds_getindex(seq, from+i-1)
-        kmer = kmer << 2 | trailing_zeros(nt)
-        isok &= iscertain(nt)
+@inline function Base.iterate(it::SpacedKmerIterator{T,S}, state=(it.start-it.increment, 1, 0, UInt64(0))
+    ) where {T,S<:Union{ReferenceSequence, BioSequence{<:FourBitNucs}}}
+    i, pos, filled, kmer = state
+    i += it.increment
+
+    while i ≤ it.stop
+        nt = reinterpret(Int8, inbounds_getindex(it.seq, i))
```
BenJWard: Should this be UInt8?

jakobnissen: Sure, it compiles to the same instructions anyway, so shouldn't matter.

jakobnissen:
> Are you finished with this @jakobnissen, or do you still have a bit more to do?
I just discovered that iterating over `eachkmer(seq, 4)` is ~700x times slower than `each(DNAKmer{4}, seq)` because the former is unable to type-specialize due to 4 being a value, not a type. It seems bad that such a performance trap awaits users, who will have little change to realize this subtle difference. What do you say we do? Rewrite `eachkmer` to take `Val{k} where {k}` or remove the `eachkmer` interface?

BenJWard:
I would vote for removing the `eachkmer` interface. But that would mean changing the API which would require a version change. I'm doing API changes over on my fork too, so a v2.x of BioSequences is in the works. It depends on if you want to merge now with `eachkmer` (which would be a minor version bump with speed improvements) and I can just remove it later as part of my more API altering mods.

jakobnissen:
Alright. I say leave it and merge it now, then. Note that with my changes, `eachkmer` is 4x slower than the existing `eachkmer` method, because it's even less certain about types than before now that there are multiple subtypes of `AbstractKmerIterator` it could return. However, with or without this PR merged, `eachkmer` is totally unsuitable for speed-sensitive code anyway, so in practice it doesn't change anything. Merge it. :)

BenJWard:
I've just added warnings that should notfiy people that eachmer is slow and is being deprecated to discourage its use. I'll just make sure it doesen't screw up the tests and I'm happy to merge it.

BenJWard:
Merged, thank you so much @jakobnissen for taking the time to do this!

kescobo:
Nice work! Sorry about the DataStructures dep, that was my bad. At one point I was trying to use some special array type that I thought would be faster. Forgot to remove it when we decided it was the wrong path. Good catch!

PR #54, opened 2019-01-08
Add unsafe_setindex for more Base.setindex methods

BenJWard:
This PR will add `unsafe_*` versions of functions that edit sequences, that do not boundscheck _nor call `orphan!` to check if the sequence in question owns it's underlying data, possibly triggering the copy on write mechanism_.

`unsafe_*` versions have currently been implemented for:

- `Base.setindex`

@jakobnissen WIP, but this is what it will look like. Will do the same in the PR for transformation functions, `copyto!` and so on and so on. So we will have a consistent way of avoiding the COW checks - at your own risk ;)


BenJWard:
Just one more thing to review before this can go in, and that the use of `copyto!` in one of the `setindex` methods.

I've the idea that it can have an `unsafe_setindex` which calls an `unsafe_copyto`. But the `copyto!` COW check is different to the check normally performed, so we need to make sure someone using this particular `unsafe_setindex` doesen't really screw up. But I'm not 100% sure there will ever be a situation in which one would want to ever want to do this. Especially since even with a COW check, a `copyto!` operation does one check and then sets many consecutive nucleotides at once, so it's already quite efficient.

jakobnissen:
I've just looked over the changes: Everything looks good.
About the `copyto!` call in the `setindex!` method, assuming you mean [this one](https://github.com/BioJulia/BioSequences.jl/blob/524af0c32bfa33b7605900534d9d45eb1a36eb08/src/longsequences/indexing.jl#L60), I've just run a few benchmarks on it. I think you're right: Since the boundschecks and COW-check only happens once, only an insignificant amount of time  is spent on the checks compared to the other operations. And I also agree that, since [this COW-check](https://github.com/BioJulia/BioSequences.jl/blob/524af0c32bfa33b7605900534d9d45eb1a36eb08/src/longsequences/copying.jl#L29) is different than the others, the user might be mislead and use `unsafe_setindex!` wrongly.
The function is ridiculously fast anyway - in the sub-microsecond domain even for hundreds of nucleotides copied - perhaps consider adding some `@inbounds` to the lines where you set and get indices in the `.data`.arrays in [this function?](https://github.com/BioJulia/BioSequences.jl/blob/524af0c32bfa33b7605900534d9d45eb1a36eb08/src/longsequences/copying.jl#L36) It's just a modest 15% speed increase, but hey, it's free. 
When looking over the code, I did find something weird, though. What exactly does [this function](https://github.com/BioJulia/BioSequences.jl/blob/524af0c32bfa33b7605900534d9d45eb1a36eb08/src/longsequences/copying.jl#L9) do? It looks broken.

BenJWard:
Awesome, I've made relevant changes, it turns out that function is broken I think, and commenting it out has absolutely no effect.

PR #55, opened 2019-01-09
Reworked random LongSequence generation

jakobnissen:
# Rework of random LongSequence generation

With BioSequences version2 coming, it's a good time to change any interfaces that could use a change. What follows is a complete rewrite of the random sequence generation - both the internals and the interface. Note that I have not yet written any tests, nor even checked that the current tests pass - please don't merge this until I have. For now, let's just discuss this proposed interface.

## Rationale:
Generating random sequences is a typical step in many bioinformatics workflows. For example, when developing a new method, I will create random DNA sequences from some simple distribution, e.g. 1% N's. The interface presented in this PR allows you to customize sequence generation a bit more, while still keeping the convenient functions like `randomdnaseq`.   

## Types of changes

Complete rewrite of randseq.jl. Furthermore, `LongSequence{A}(len::Integer)` now throws an ArgumentError if `len` < 0, instead of creating a zero-length sequence.

## The new interface

The simplest way to generate sequences is still the functions `randdnaseq`, `randrnaseq` and `randaaseq`. While their internals have changed, their behaviour remains the same.

__Generation sequences from a default distribution: `randseq(A::Alphabet, len::Interger)`__
This is the main method for creation of random LongSequence of a specific alphabet from a default distribution. User-defined alphabets should overwrite this method. Compared the e.g. `randdnaseq`, the advantage of this is that you can more easily and effectively generate for example a `LongSequence{DNAAlphabet{2}}`.

Specialized methods have been implemented for `A::NucleicAcidAlphabet{2}`, `A::NucleicAcidAlphabet{4}` and `AminoAcidAlphabet()`. The default fallback (e.g. for a user-defined alphabet A) is a uniform distribution over all `symbols(A)`.

`randdnaseq`, `randrnaseq` and `randaaseq` calls this method internally.

__Specifying custom distributions__
To do this, the user needs to create either a `SamplerUniform` or a `SamplerWeighted`, both subtypes of `Random.Sampler`. For example:

`julia> sampler = SamplerUniform(dna"ACGTN");`

This sampler can then be queried for random values by `rand(sampler)`. Similarly, for non-uniform distributions, a user might do:

```
julia> sampler = SamplerWeighted(dna"ACGTN", fill(0.2475, 4))
julia> rand(sampler)
DNA_G
```

__Generating sequences from a custom distribution__
To do this, simply pass the sampler to `randseq`:

```
julia> seq = randseq(DNAAlphabet{4}(), sampler, 100)
100nt DNA Sequence:
TGCTCTACAACGGAAACAACCTATAACACATAACTGCTA...GTGTACGGACATATTGANTCATGTGATTAACAGTAGATC
```

## Notes
* I'm not entirely sure I've correctly understood when to use an alphabet type and when to use the singleton instance of it. All methods I've written uses instances.
* Feedback on the interface welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- New subtypes of Random.Sampler, SamplerUniform and SamplerWeighted.
+- Random `LongSequence`s can now be created with `randseq`,
+  optionally using a sampler to specify element distribution.
+- All random `LongSequence` generator methods take an optional AbstractRNG argument.
+- Add methods to `randseq` to optimize random generation of `NucleicAcid` or `AminoAcid` `LongSequence`s.
+
 ### Changed
 - The abstract `Sequence` type is now called `BioSequence{A}`.
 - The type previously called `BioSequence{A}` is now `LongSequence{A}`.
@@ -49,7 +56,7 @@ is less than or equal to zero.
 ### Added
 - Position weight matrix search functionality.
 - A generalised composition method.
-- `typemin` and `typemax` methods for `Kmer` types. 
+- `typemin` and `typemax` methods for `Kmer` types.
 
 ### Changed
 - `MinHash` function now generalised to `Reader` types.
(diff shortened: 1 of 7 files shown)


jakobnissen:
Pinging @BenJWard since this depends on, and has influence on, his current work on BioSequences.jl version2 interface.

jakobnissen:
@BenJWard This can be merged as soon as checks pass and you're satisfied with the content.

BenJWard:
Thanks for doing this @jakobnissen!

PR #57, opened 2019-01-12
SkipmerFactory

BenJWard:
✨ A new iterator that we can dispatch on in addition to @jakobnissen's faster iterators over Kmers. 

The idea being that when someone iterates over Skipmer types that are infact Kmers (Skipmer types where parameters M == N == 1), @jakobnissen's methods are chosen during dispatch, otherwise, this iterator is chosen during dispatch.

- [ ] :art: The changes implemented is consistent with the [julia style guide](https://docs.julialang.org/en/stable/manual/style-guide/).
- [ ] :blue_book: I have updated and added relevant docstrings, in a manner consistent with the [documentation styleguide](https://docs.julialang.org/en/stable/manual/documentation/).
- [ ] :ok: There are unit tests that cover the code changes I have made.


BenJWard:
Not quite ready yet, as I have to catch this series of commits up with the other changes made for v2 and make sure everything still works :+1:.

BenJWard:
@jakobnissen This is basically working now. If you have any comments let me know.

I need to add tests to check dispatch between this iterator, and your more specialised iterators happens correctly for different Skipmer/Kmer types.

But you can see how it produces all the skipmers from a sequence in a single pass, by constructing - for example - the first, second, and third skipmer of a single cycle, (for skipmers with a cycle length of 3) at the same time, and returning them when they are finished, at each position.

jakobnissen:
I've had a look at the code, and I have no comments. Mostly because I find it a little hard to comprehend: There are very few comments or docstrings, and many of the variable names are unclear to me - of course, this does not mean that I find the code bad, just impenetrable.
Iterating over Skipmers doesn't work for me - more specifically, the `reinterpret` at eachskipmer.jl ln 144 doesn't work since `Skipmer` is not a primitive type.
But certainly interesting - I'm looking forward to seeing it in action. How fast is iterating Skipmers over an e.g. `DNAAlphabet{4}` sequence compared to iterating kmers (i.e. using specialized code for then the skipmer is a kmer)?

BenJWard:
It is difficult to visualise, I'll add more comments to clean up the code, it was based on my supervisors C++ code which is rarely pleasant to read! 

The iterator basically work's like this:
Say M and N of the skipmer is set to 2 and 3, and K = 14

Then the skipmers would sample 2 nucleotides in every 3:

```
"GAGGGGGATGCCCCTCTTTGAGCCCAAGG" - 29 bases
"GA GG GA GC CC CT TG"
```

Then move along 1 bp, and do it again for the next kmer:

```
"GAGGGGGATGCCCCTCTTTGAGCCCAAGG" - 29 bases
" AG GG AT CC CT TT GA"
```

And so on. Given the N parameter is 3. There are three "phases" you can represent as 3 masks:
```
 12345678901234567890123456789
 xx xx xx xx xx xx xx xx xx xx
  xx xx xx xx xx xx xx xx xx x
 x xx xx xx xx xx xx xx xx xx 
"GAGGGGGATGCCCCTCTTTGAGCCCAAGG" - 29 bases
```
From this you can see an efficient iterator, going base by base in the ref sequence, is able too build the next kmer in each of the three "phases" at the same time. e.g. when the iterator touches base 1, it needs to add that base to the next skipmer in phase 1, and the next skipmer in phase 3 (both masks cover that position). So rather than requiring three passes of the reference sequence to generate all the skipmers, this iterator can do it in 1 pass.

So the full picture looks something like this:
```
12345678901234567890123456789
 xx xx xx xx xx xx xx xx xx xx
"GAGGGGGATGCCCCTCTTTGAGCCCAAGG" - 29 bases
"GA GG GA GC CC CT TG         " - 1          "CAAGGGGCTCCCTC"          
"   GG GA GC CC CT TG GC      " - 4
"      GA GC CC CT TG GC CA   " - 7
"         GC CC CT TG GC CA GG" - 10

 12345678901234567890123456789
  xx xx xx xx xx xx xx xx xx x
"GAGGGGGATGCCCCTCTTTGAGCCCAAGG" - 29 bases
" AG GG AT CC CT TT GA        " - 2              
"    GG AT CC CT TT GA CC     " - 5
"       AT CC CT TT GA CC AA  " - 8

 12345678901234567890123456789
 x xx xx xx xx xx xx xx xx xx 
"GAGGGGGATGCCCCTCTTTGAGCCCAAGG" - 29 bases
"  GG GG TG CC TC TT AG       " - 3
"     GG TG CC TC TT AG CC    " - 6
"        TG CC TC TT AG CC AG " - 9
```

jakobnissen:
Ah yes, now I see! It all looks quite neat.
It's certainly much more complicated to calculate skipmers than kmers - just like calculating kmers with ambiguous nucleotides is much more complicated than for 2-bit nucleotides, I suppose.
Why is it you keep track of both the forward and the reverse kmer - isn't it much more efficient to simply reverse-complement it after it's been constructed? Or will that not work, because the positions of the forward and the reverse kmer will be different?

e.g, like this, where the first fw skipmer is not the same as the last rv skipmer:
```
fw:   xx xx xx -> xx xx xx
   5' TAGTAGATCGCATGCCTGTAGA 3'
   3' ATCATCATGCGTACGGACATCT 5'
rv:     xx xx xx <- xx xx xx
```
However, if so, this is also the case for kmers if the step is not 1. For kmers, however, we just usually take the positions for one read direction and reverse-complement. Is this correctly understood?

BenJWard:
> Or will that not work, because the positions of the forward and the reverse kmer will be different?
> 
> e.g, like this, where the first fw skipmer is not the same as the last rv skipmer:
> ```
> fw:   xx xx xx -> xx xx xx
>    5' TAGTAGATCGCATGCCTGTAGA 3'
>    3' ATCATCATGCGTACGGACATCT 5'
> rv:     xx xx xx <- xx xx xx
> ```
> However, if so, this is also the case for kmers if the step is not 1. For kmers, however, we just usually take the positions for one read direction and reverse-complement. Is this correctly understood?

That's a really good question. I just had to have a chat with my team about this.

The correct way to think about the forward skipmer and its reverse is more like this diagram:

```
fw:   TA TA AT -> TG CT TA
   5' TAGTAGATCGCATGCCTGTAGA 3'
      xx xx xx .. xx xx xx
   3' ATCATCTAGCGTACGGACATCT 5'
rv:   AT AT TA <- AC GA AT

fw: TATAATTGCTTA
rv: TAAGCAATTATA
```
This is because in the Skipmer paper you'll see, and in the code, skipmers have a property called the span (S in the image below):

image

Skip-mers map to k-mers in a transformed space and so the rc needs to be the rc of the “equivalent k-mer”, as my supervisor would put it.

Or more straightforwardly the rc of a given skipmer is the skipmer taken on the opposite strand that spans the same region (or nucleotides)."

So this scheme of producing RC'd skipmers would be wrong:

```
 fw:   xx xx xx -> xx xx xx
>    5' TAGTAGATCGCATGCCTGTAGA 3'
>    3' ATCATCATGCGTACGGACATCT 5'
> rv:     xx xx xx <- xx xx xx
```

Precisely because the two skipmers span slightly different regions of the molecule, and so you're taking different bases, and the information contained in the fw and bw skipmers would not be the same information.
```

**Comment by BenJWard** [comment]
@jakobnissen 

As for why the rv is constructed at the same time. We did that with the rationale of it being faster. But I'm not 100% convinced this is true in the julia iterator... it may have been for my supervisors C++ code if it used strings or something. I've been thinking of a generic `EachCanonical` iterator that wraps any iterable source of `BioSequences`, and returns the canonical form of the sequence each time. Which is more elegant from a design perspective: kmer and skipmer iterators return the skipmers, as they are, and if you want the canonical ones, then you can wrap those iterators in the `EachCanonical` one. But we'd need to implement this and benchmark it, as I don't know how the performance is going to look. I can see an argument as to why building the rv as you go might be performant, but I can also see an reason why it might not be significant (or even true) for BioJulia.

**Comment by BenJWard** [comment]
@jakobnissen To RC a fw skiper after the iterator produced it, would require the following ops:
```
@inline function reversebits(x::UInt64, ::BitsPerSymbol{2})
     x =  ((x >> 2) & 0x3333333333333333) | ((x & 0x3333333333333333) << 2)
     x = reversebits(x, BitsPerSymbol{4}())
     return x
end

@inline function reversebits(x::UInt64, ::BitsPerSymbol{4})
     x = ((x >> 4 ) & 0x0F0F0F0F0F0F0F0F) | ((x & 0x0F0F0F0F0F0F0F0F) << 4 )
     x = ((x >> 8 ) & 0x00FF00FF00FF00FF) | ((x & 0x00FF00FF00FF00FF) << 8 )
     x = ((x >> 16) & 0x0000FFFF0000FFFF) | ((x & 0x0000FFFF0000FFFF) << 16)
     x = ((x >> 32) & 0x00000000FFFFFFFF) | ((x & 0x00000000FFFFFFFF) << 32)
     return x
end

function Base.reverse(x::T) where {T <: Skipmer}
    bits = encoded_data(x)
    rbits = reversebits(bits, BitsPerSymbol{2}())
    return T(rbits >> (sizeof(bits) * 8 - 2 * length(x)))
end
``` 
Which is quite a number of ops per iteration to make an RC of a skipmer.

Building the rv skipmer as you go, with each iteration of the iterator, requires:
```
rbits = ~fbits & 0x03
it.rkmer[ni] = (it.rkmer[ni] >> 2) | (UT(rbits) << firstoffset(it))
```

Which is a lot less ops to do. BUT depending on optimisations and compilation... I dunno how much faster it is.

BenJWard:
The iterator has been replaced with a SkipmerFactory that can be loaded with a sequence, and calling `nextmer` will give a (position, fwmer, bwmer) tuple. It also has `Base.iterate` methods defined for it.

BenJWard:
Ok at long last I am happy enough with this PR to merge it, the SkipmerFactory design is down now, any further perf improvements or fixes will be in subsequent PRs, although I have spent a long time making sure this addition is correct.

PR #63, opened 2019-05-10
Project TOML file

BenJWard:
This PR removes any dependencies no longer used by the BioSequences module, and will replace the REQUIRE file with a Project.toml


cecileane:
Do you have plans to merge this version into master, so that projects using `BioSequences` can also use their own `Project.toml` and register new versions on the general registry?

BenJWard:
I was going to leave master as it is until the version2 branch is in a state I am happy with, at which point master will be updated an a V2.X of BioSequences, with a new API and sequence types will be released. The work for BioSequences v2.0 is not far from completion now: one or two weeks should do it with the time I have available.

As far as I can tell, BioSequences 1.1 using REQUIRE hasn't been causing any issues, I've been developing and releasing packages that use Project.toml and BioSequences without an issue so far.

cecileane:
Oh it makes complete sense.

I was under the impression that "A package with a project file cannot depend on one without a project file" (from [here](https://docs.julialang.org/en/v1/manual/code-loading/#Environments-1)). But this may not be a problem for packages that were in METADATA before the switch to the General registry, as all these packages are in there, and with their own uuid. Sorry for the trouble.

Looking forward to BioSequences v2.0!

PR #79, opened 2020-01-06
Faster and more general kmer reverting

jakobnissen:
Basically this is a small PR with two minor changes:

* `reversebits` now use `bswap` in order to be faster.
* `reversebits` is now written to be a bit more generic, as it now works with UInt8, UInt16, UInt32, Uint64 and UInt128.

### DIFF
--- a/src/bit-manipulation/bit-manipulation.jl
+++ b/src/bit-manipulation/bit-manipulation.jl
@@ -2,33 +2,16 @@
 include("bitindex.jl")
 include("bitpar-compiler.jl")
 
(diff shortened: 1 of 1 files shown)


jakobnissen:
Looks like some documentation tests failed. I'm not sure what that means - there was no documentation on these internal function before, either.

BenJWard:
I wouldn't worry about the docs for now. It's probably something I'll migrate to GitHub actions and so on before the next release anyway.

BenJWard:
Thanks @jakobnissen!

PR #81, opened 2020-01-10
Fast ASCII String to LongSequence conversion

jakobnissen:
# Fast ASCII String to LongSequence conversion

This PR makes it so that if a LongSequence is constructed from a String, or `encode_copy!`'ed from a String, it detects if the String is an ASCII string. If so, it wraps a byte vector around it and uses the faster vector to LongSequence conversion path.

Edit: Should probably post some benchmarks. Here, I run `LongSequence{DNAAlphabet{4}}(st)` for varying lengths of string:
```
Len   Before     After (time in ns)
    0     48        62
    1     60        75
   50    635       134
  150   1863       294
 1000  11630      1164
10000 115815     10220
```

### DIFF
--- a/src/longsequences/constructors.jl
+++ b/src/longsequences/constructors.jl
@@ -1,6 +1,6 @@
 ###
 ### Constructors
-### 
+###
 ###
 ### Constructor methods for LongSequences.
 ###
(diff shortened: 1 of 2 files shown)


BenJWard:
Can this be rebased onto master? Master has all the version2 branch changes now and some more, since v2 got released to the BioJulia registry.

jakobnissen:
Yes sure! I didn't know master had been updated. I'm currently trying to figure out how to generalize it to all alphabets with the trait BioSequences.AsciiAlphabet. But that's probably better left for another PR.

jakobnissen:
Superseded by #83, which is larger in scope.

Issue #82, opened 2020-01-10
Implement fast String to LongSequence conversion for more alphabets


jakobnissen:
In #81 , I made LongSequence conversion of ASCII strings much faster by using an already-optimized method for conversion to 4-bit nucleotide alphabets. However, there is no corresponding fast method for 2-bit alphabets nor amino acid alphabets, though there easily could be.
That should be implemented some time.

Edit: Instead of current dispatch model, for instantiating a LongSequence of alphabet type DNAAlphabet, RNAAlphabet and AminoAcidAlphabet, don't even check for ASCII. Just wrap a vector around it always, since non-ascii strings will not be valid anyway. That saves the call to isascii, which is O(n).

jakobnissen:
Implemented by #83

PR #83, opened 2020-01-13
More efficient ASCII/LongSequence conversion

jakobnissen:
# Refactoring/rewrite of ASCII/Longsequence conversion

This PR supersedes #81 and resolves issue #82. It is a change in the internals of BioSequence to gain more performance. Behaviour should not be changed by this PR.

In BioSequences, a biological symbol is represented in three different format:
* As a specialized type, e.g `DNA_G`
* As a Char, e.g. `'G'`
* As the encoded bits in an alphabet, e.g. `0x02` for `RNAAlphabet{2}`

BioSequences + BioSymbols provide efficient, generic conversions between these three. Some alphabets, however, including the five most used ones, have a further property: Their `Char` encodings are all ASCII, which means they have another representation:
* UInt8 represention, e.g. `UInt8('G')`, which is also their `String` codeunit representation.

For these alphabets, this representation can replace `Char`, and we can rely on base Julia to convert efficiently between UInt8 and Char. This property enables much more efficient String/LongSequence/byte operations for these alphabets, but it is not used by BioSequences.

This PR is an attempt to provide a coherent, *internal* (NOT an API) interface for efficient conversions to and from the UInt8 representation.

The interface will be generic in the sense that user-defined alphabets need only implement a few specific methods to get full performance. This is implemented via the `codetype` trait.

### But why?
Speed. Also, it gives developers some kind of guarantee that they can assume ASCII symbols for some alphabets, which can make things much easier.

Here's a short benchmark showing pretty big improvements. This is the time taken to instantiate a `LongSequence` from a `String`. (updated 2020-01-14):
```
2-bit nucleotides
Length  before   after
     0      51      51
     1      61      53
    50     651      93
   150    1947     158
  1000   12108     629
 10000  119868    5944
100000 1204272   56989

4-bit nucleotides
Length  before   after
     0      51      51
     1      75      55
    50    1302      87
   150    3703     151
  1000   24654     654
 10000  244861    6128
100000 2510870   78783

amino acids
Length  before   after
     0      51      50
     1      64      53
    50     825      84
   150    2429     149
  1000   15396     665
 10000  153504    6546
100000 1581427   99056
```
There are also performance benefits for `encode_copy!` of strings and `Vector{UInt8}`, though they are smaller.

### Things to review
Let's look through the possible conversions:
* `UInt8` -> symbol: Not implemented, probably not needed
* `UInt8` -> encoding: Implemented by this PR
* `UInt8` -> `Char`: Implemented efficiently in Base
* symbol -> `UInt8`: Implemented by `stringbyte` function in #80 
* symbol -> encoding: Efficiently implemented in by `encode`
* symbol -> `Char`: NOT efficiently implemented, could use a PR.
* encoding ->`UInt8`: Not efficiently implemented. Perhaps of use for speeding up String conversion?
* encoding -> symbol: Efficiently implemented by `decode`
* encoding -> `Char`: Not efficiently implemented, but not needed.
* `Char` -> `UInt8`: Efficiently implemented in Base
* `Char` -> symbol: Efficiently implemented in BioSymbols
* `Char` -> encoding: Not efficienly implemented, does a `Char` -> symbol -> encoding. Probably not needed to be efficient.

So to me it seems only symbol -> `Char` and encoding -> `UInt8` are interesting to look at.

### To do
* ~~Add docstring to `AsciiAlphabet` to tell what user needs to define to make it work.~~
* ~~Make symbol -> Char more efficient in BioSymbols.~~
* ~~Possibly implement encoding -> UInt8 conversion for even faster conversion to `String`~~ edit: This turns out to be relatively hard to do, and will take maybe 100 lines of code. Not worth the complexity, IMO. Parsing from Strings is more important than parsing to Strings.
* ~~Review tests again to make sure everything new done by this PR is covered in the test cases.~~

### DIFF
--- a/src/alphabet.jl
+++ b/src/alphabet.jl
@@ -8,6 +8,8 @@
 ### License is MIT: https://github.com/BioJulia/BioSequences.jl/blob/master/LICENSE.md
 
 """
+# Alphabets of biological symbols.
+
 `Alphabet` is perhaps the most important type trait for biological sequences in
 BioSequences.jl.
 
@@ -98,7 +100,7 @@ for alph in (DNAAlphabet, RNAAlphabet)
 end
 
 minimal_alphabet(::Type{A}) where A <: DNAAlphabet = DNAAlphabet{2}
-minimal_alphabet(::Type{A}) where A <: RNAAlphabet = RNAAlphabet{2} 
+minimal_alphabet(::Type{A}) where A <: RNAAlphabet = RNAAlphabet{2}
 minimal_alphabet(x::A) where A <: NucleicAcidAlphabet = minimal_alphabet(typeof(x))
 
 ###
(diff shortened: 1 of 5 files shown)


BenJWard:
Is symbol -> Char. Not handles by BioSymbols?

jakobnissen:
Yep, but it can be made more efficient. Let me know if you think I'm overengineering this stuff. https://github.com/BioJulia/BioSymbols.jl/pull/36

jakobnissen:
I'm satisfied with this PR now, and it looks like code coverage is OK. It is ready for merging.

PR #84, opened 2020-01-14
Faster reverse! of LongSequence

jakobnissen:
# Faster reverse! of LongSequence

Another speed-related internal change. So, currently `reverse` was up to 100x faster than `reverse!` because the former used neat bit-tricks that the latter didn't. So, why didn't the latter do it? Because the bit tricks only worked when copying the underlying data vector.

This PR adds new tricks to allow `reverse!` to be fast, too. See benchmarks at the bottom.

### EDIT: With newest changes, this version is faster than old one. No decision anymore.
**Here's the decision:** With this PR there are two similar bit-hacks for speeding up reverse-complementation. It would be nice to discard the old `reverse` and instead implement reverse simply as `_reverse!(copy(seq), BitsPerSymbol(seq))`. _But this would be slower than the old_ `reverse`, because `copy(seq)` is slow.

Here's a benchmark of what happens **if I discard the old `reverse`**. The regression is about 20-35%. However, if we allow `reverse` to call `orphan!`, the speed can be kept without keeping the old, more complex code. This will cause an unneeded copy in some cases.
```
Inplace reverse!
Type   Len    Before    After
2-bit  0         10       11
2-bit  1         10       15 
2-bit  32       394       14
2-bit  500     5928       38
2-bit  5000   58627      198
4-bit  0         10       11
4-bit  1         10       15
4-bit  32       359       20
4-bit  500     5343       49
4-bit  5000   53449      324
AA     0         12       11
AA     1         12       14
AA     32       367       20
AA     500     5455       76
AA     5000   54672      229

Copying reverse
Type   Len    Before    After
2-bit  0         37       44
2-bit  1         40       49 
2-bit  32        40       48
2-bit  500       70       82
2-bit  5000     338      402
4-bit  0         37       43
4-bit  1         39       47
4-bit  32        40       50
4-bit  500       83      102
4-bit  5000     527      708
AA     0         54       48
AA     1         58       53
AA     32       418       60
AA     500     5572      148
AA     5000   54844     1054 
```

So, is it worth keeping the old `reverse` for a 20-35% faster reversion, or should it be discarded to keep the code complexity lower?

### DIFF
--- a/src/longsequences/transformations.jl
+++ b/src/longsequences/transformations.jl
@@ -59,22 +59,62 @@ end
 
 Reverse a biological sequence `seq` in place.
 """
-function Base.reverse!(seq::LongSequence)
-    orphan!(seq) # TODO: Is the orphan call really nessecery given the indexing calls will also call orphan?
-    @inbounds for i in 1:div(lastindex(seq), 2)
-	    x = seq[i]
-        i′ = lastindex(seq) - i + 1 
-	    seq[i] = seq[i′]
-	    seq[i′] = x
+Base.reverse!(seq::LongSequence{<:Alphabet}) = _reverse!(orphan!(seq), BitsPerSymbol(seq))
+
+function _reverse!(seq::LongSequence{A}, ::BitsPerSymbol) where {A <: Alphabet}
+    i, j = 1, lastindex(seq)
+    while i < j
+        seq[i], seq[j] = seq[j], seq[i]
+        i += 1
+        j -= 1
     end
     return seq
 end
 
+function _reverse!(seq::LongSequence{A}, ::B) where {A <: Alphabet,
+    B <: Union{BitsPerSymbol{2}, BitsPerSymbol{4}, BitsPerSymbol{8}}}
+
+    # Reverse order of chunks and bits in chunks in one pass
+    data = seq.data
+    len = length(data)
+    bps = BitsPerSymbol(seq)
+    @inbounds for i in 1:len >>> 1
+        data[i], data[len-i+1] = reversebits(data[len-i+1], bps), reversebits(data[i], bps)
+    end
+    @inbounds if isodd(len)
+        data[len >>> 1 + 1] = reversebits(data[len >>> 1 + 1], bps)
+    end
+
+    # Reversion of chunk bits may have left-shifted data in chunks, so we must
+    # shift them back to an offset of zero
+    lshift = offset(lastbitindex(seq) + bits_per_symbol(A()))
+    rshift = 64 - lshift
+    if !iszero(lshift)
+        @inbounds for i in 1:len-1
+            left = seq.data[i] >>> (rshift & 63)
+            right = seq.data[i+1] << (lshift & 63)
+            seq.data[i] = left | right
+        end
+        @inbounds seq.data[len] >>>= rshift
+    end
+
+    return seq
+end
+
+"""
+    reverse!(seq::LongSequence)
+
+Reverse a biological sequence.
+"""
+function Base.reverse(seq::LongSequence{A}) where {A<:Alphabet}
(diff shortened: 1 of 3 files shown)


src/longsequences/transformations.jl
```diff
@@ -59,22 +59,62 @@ end
...
-    @inbounds for i in 1:div(lastindex(seq), 2)
-	    x = seq[i]
-        i′ = lastindex(seq) - i + 1 
-	    seq[i] = seq[i′]
-	    seq[i′] = x
+Base.reverse!(seq::LongSequence{<:Alphabet}) = _reverse!(orphan!(seq), BitsPerSymbol(seq))
+
+function _reverse!(seq::LongSequence{A}, ::BitsPerSymbol) where {A <: Alphabet}
+    i, j = 1, lastindex(seq)
+    while i < j
```
BenJWard: Like the generic fallback kernel. Can we make this a bit faster by using `@inbounds`?

src/longsequences/transformations.jl
```diff
@@ -59,22 +59,62 @@ end
...
     return seq
 end
 
+function _reverse!(seq::LongSequence{A}, ::B) where {A <: Alphabet,
+    B <: Union{BitsPerSymbol{2}, BitsPerSymbol{4}, BitsPerSymbol{8}}}
+
+    # Reverse order of chunks and bits in chunks in one pass
+    data = seq.data
+    len = length(data)
+    bps = BitsPerSymbol(seq)
```
BenJWard: Can this be removed by giving `::B` a name in the function head?

src/longsequences/transformations.jl
```diff
@@ -59,22 +59,62 @@ end
...
+    @inbounds for i in 1:len >>> 1
+        data[i], data[len-i+1] = reversebits(data[len-i+1], bps), reversebits(data[i], bps)
```
BenJWard: Does it make a difference if you split this into multiple lines? I'm curious is julia would do any tuple expanding.... stuff, that adds to time.

BenJWard:
So to make sure I understand correctly: sub-seq-ing the entire length of an input sequence, forcefully orphaning it, and then doing the `_reverse!`, is faster (yet basically has the same effect) than doing `_reverse!(copy(seq))`? Is there anything we can do to improve `copy` so as the trick is not nessecery?

src/longsequences/transformations.jl
jakobnissen: Yes, that should happen.

src/longsequences/transformations.jl
jakobnissen: Definitely.

src/longsequences/transformations.jl
jakobnissen: Looks like they compile to identical native code for me. Actually this part is surprisingly efficient.

jakobnissen:
Turns out that even after merging #85 , using the old, specialized method is still 10% faster for long sequences of `NucleotideAlphabet`. I think it's not worth the extra method and advocate for deleting the old, specialized method, but I'll leave it up to you.
```
Length  before   after
     0      40      37
     1      42      40
    50      48      45
   150      59      55
  1000     141     153
 10000    1170    1305
100000   27619   29986
```

jakobnissen:
Figured out how to SIMD parallelize the second loop in `_reverse!`, and inlined the function. Finally, this version is faster than the old `reverse`, which I have now removed. All tests pass. This PR is ready to be merged.

BenJWard:
Aw GitHub, y u no gif?

BenJWard:
This is brilliant @jakobnissen, thanks for taking the time to do this... I'm having flashbacks to trying to simd optimize the loops in the bitparallel counting stuff.

PR #87, opened 2020-01-16
Add copy! method to LongSequence

jakobnissen:
# Add copy! method to LongSequence

The method is intended to be used for something like this:
```
seq = LongSequence{DNAAlphabet{2}}()
for line in eachline(file)
    copy!(seq, line)
    # process seq somehow
end
```

This is highly efficient, not reallocating the `LongSequence`. Note that this can potentially be useful in FASTX, since you can also `copy!` views of the underlying FASTA data.

The `copy!(dst, src)` methods works like arrays, copying to the dst and resizing it. Its implementation is effectively:
```
function copy!(dst, src)
    resize!(dst, length(src))
    copyto!(dst, src)
end
```

* For LongSequence -> LongSequence of same alphabet, it simply calls `resize!`, then inbounds `copyto!`.
* For String -> ASCII alphabet LongSequence, it avoids calling the O(n) `length(string)` by wrapping it in a byte vector, then using `resize!` and `encode_chunks!`.
* For any other source -> LongSequence, it calls `resize!`, then `encode_copy!`

### DIFF
--- a/src/longsequences/copying.jl
+++ b/src/longsequences/copying.jl
@@ -54,6 +54,12 @@ end
     return dst
 end
 
+function Base.copy!(dst::LongSequence{A}, src::LongSequence{A}) where {A <: Alphabet}
+    len = length(src)
+    resize!(dst, len)
+    @inbounds copyto!(dst, 1, src, 1, len)
+end
+
 # Dispatch to alphabet type
 function Base.copy!(seq::LongSequence{A}, src) where {A<:Alphabet}
    return copy!(seq, src, codetype(A()))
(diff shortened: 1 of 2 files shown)


BenJWard:
I think this should be copyto! instead of copy! It would be more consistent with FASTX, inc which you can `copyto!` a `FASTA.Record`s or `FASTQ.Record`s sequence data into a `BioSequence`.

BenJWard:
Actually, no, reading the docs, the semantics of `copy!` is correct as you've done here. I'll have to check FASTX to make sure its consistent.

PR #88, opened 2020-01-16
Fix orphan! bug

jakobnissen:
# Fix orphan! bug

This PR makes four changes:
### 1: Fix bug in orphan!
Currently on `master`:
```
julia> a =  BioSequences.orphan!(LongDNASeq("A"^17), 16, true)
17nt DNA Sequence:
AAAAAAAAAAAAAAAA-

julia> a.data
1-element Array{UInt64,1}:
 0x1111111111111111
```
This is fixed in this branch

### 2: Optimize orphan!
Because why not, it's used all the time internally. It's especially important that it's fast when the sequence is not shared (as is likely the case most of the time). See benchmarks at bottom.

### 3: Optimize `length(::LongSequence)`
Current imlpementation checks for overflows, which is only possible if the `part` field contains negative numbers. But this can never happen, and if it does, it causes all kinds of bugs anyway (like `eachindex` not agreeing with `firstindex` etc). It also causes an annoying branch that actually can cause significant slowdown.

### 4: Optimize `resize!`
Make it do nothing if size does not change, and then use the newly-introduces `_orphan!` to more effectively do the resize if necessary.

### Benchmarks
```
Orphan! non-shared seq
Len     Before    After
0       6.8       1.9  
1       7.2       1.7
15      7.1       1.7
32      7.1       1.7
500     7.0       1.9
50000   7.1       1.7

Force orphan! non-shared seq
Len     Before    After
0       31.5      29.0
1       35.2      32.0
15      35.2      31.8
32      36.0      32.8
500     52.5      47.9
50000 1816.0    1544.0

orphan! shared seq
Len     Before    After
0       41.8      36.9
1       45.0      40.8
15      45.2      39.8
32      46.0      40.6
500     63.0      56.4
50000 1852.0    1574.0
```

### DIFF
--- a/src/longsequences/longsequence.jl
+++ b/src/longsequences/longsequence.jl
@@ -90,12 +90,11 @@ function orphan!(seq::LongSequence,
 	if !seq.shared & !force
 	    return seq
 	end
-	return _orphan!(seq, size, force)
+	return _orphan!(seq, size)
 end
 
 function _orphan!(seq::LongSequence{A},
-		 size::Integer = length(seq),
-		 force::Bool = false) where {A}
+		 size::Integer = length(seq)) where {A}
 
     j, r = bitindex(seq, 1)
     data = Vector{UInt64}(undef, seq_data_len(A, size))
--- a/src/longsequences/longsequence.jl
+++ b/src/longsequences/longsequence.jl
@@ -68,7 +68,7 @@ const LongCharSeq      = LongSequence{CharAlphabet}
 "Gets the alphabet encoding of a given BioSequence."
 BioSymbols.alphabet(::Type{LongSequence{A}}) where {A} = alphabet(A)
 Alphabet(::Type{LongSequence{A}}) where {A <: Alphabet} = A()
-Base.length(seq::LongSequence) = length(seq.part)
+Base.length(seq::LongSequence) = last(seq.part) - first(seq.part) + 1
 bindata(seq::LongSequence) = seq.data
 Base.eltype(::Type{LongSequence{A}}) where {A} = eltype(A)
 
@@ -84,25 +84,30 @@ end
 # Replace a LongSequence's data with a copy, copying only what's needed.
 # The user should never need to call this, as it has no outward effect on the
 # sequence.
-function orphan!(seq::LongSequence{A},
+function orphan!(seq::LongSequence,
+		 size::Integer = length(seq),
+		 force::Bool = false)
+	if !seq.shared & !force
+	    return seq
+	end
+	return _orphan!(seq, size, force)
+end
+
+function _orphan!(seq::LongSequence{A},
 		 size::Integer = length(seq),
 		 force::Bool = false) where {A}
-
-    if !seq.shared && !force
-        return seq
-    end
 
     j, r = bitindex(seq, 1)
     data = Vector{UInt64}(undef, seq_data_len(A, size))
 
-    if !isempty(seq) && !isempty(data)
+    @inbounds if !isempty(seq) & !isempty(data)
         x = seq.data[j] >> r
         m = index(bitindex(seq, lastindex(seq))) - j + 1
         l = min(lastindex(data), m)
-        @inbounds @simd for i in 1:l-1
+        @simd for i in 1:l-1
             y = seq.data[j + i]
             data[i] = x | y << (64 - r)
-            x = y >> r
+            x = y >> (r & 63)
         end
         if m <= l
             data[l] = x
@@ -113,7 +118,7 @@ function orphan!(seq::LongSequence{A},
     end
 
     seq.data = data
-    seq.part = 1:length(seq)
+    seq.part = 1:size
     seq.shared = false
     return seq
 end
--- a/src/longsequences/transformations.jl
+++ b/src/longsequences/transformations.jl
@@ -8,13 +8,13 @@
 Resize a biological sequence `seq`, to a given `size`.
 """
 function Base.resize!(seq::LongSequence{A}, size::Integer) where {A}
-    if size < 0
+    if size == length(seq)
+        return seq
+    elseif size < 0
         throw(ArgumentError("size must be non-negative"))
+    else
+        return _orphan!(seq, size)
     end
-    orphan!(seq, size)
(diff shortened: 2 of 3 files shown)


src/longsequences/longsequence.jl
```diff
@@ -84,25 +84,29 @@ end
 # Replace a LongSequence's data with a copy, copying only what's needed.
 # The user should never need to call this, as it has no outward effect on the
 # sequence.
-function orphan!(seq::LongSequence{A},
+function orphan!(seq::LongSequence,
 		 size::Integer = length(seq),
-		 force::Bool = false) where {A}
+		 force::Bool = false)
+	if !seq.shared & !force
```
BenJWard: Does && introduce branching vs &?

jakobnissen: Yes, `a && b` guarantees that b will not be evaluated unless a (I guess unless the compiler can prove that evaluation a has no side effects).

src/longsequences/longsequence.jl
```diff
@@ -84,25 +84,29 @@ end
...
+    @inbounds if !isempty(seq) & !isempty(data)
         x = seq.data[j] >> r
         m = index(bitindex(seq, lastindex(seq))) - j + 1
         l = min(lastindex(data), m)
-        @inbounds @simd for i in 1:l-1
+        @simd for i in 1:l-1
             y = seq.data[j + i]
             data[i] = x | y << (64 - r)
-            x = y >> r
+            x = y >> (r & 63)
```
BenJWard: This line was the source of the bug?

jakobnissen: No, actually it was just `seq.part = 1:length(seq)` instead of `seq.part = 1:size` at the end. The `& 63` just makes the bitshift compile to a single instruction. (It's needed because native bitshifts and Julia's bitshifts behave slightly different).

jakobnissen: To expand: In Julia, `x << 64` zeros x, whereas natively, it does nothing (since only the 6 lower bits of the bitshift integer actually does anything). That's also why I can't mask the left-shift of `y` with 63. `y` needs to be able to be zero'd.

BenJWard: I'm not sure I understand:

```julia
julia> @code_native 0xFFFFFFFFFFFFFFFF >> 64
	.section	__TEXT,__text,regular,pure_instructions
; ┌ @ int.jl:448 within `>>'
; │┌ @ int.jl:448 within `<='
	xorl	%ecx, %ecx
; │└
; │ @ int.jl:448 within `>>' @ int.jl:442
	cmpq	$63, %rsi
	shrxq	%rsi, %rdi, %rdx
	cmovaq	%rcx, %rdx
; │ @ int.jl:448 within `>>'
; │┌ @ int.jl:51 within `-'
	movq	%rsi, %rax
	negq	%rax
; │└
; │┌ @ int.jl:443 within `<<'
	cmpq	$63, %rax
	shlxq	%rax, %rdi, %rax
	cmovaq	%rcx, %rax
; │└
; │┌ @ int.jl:428 within `<='
	testq	%rsi, %rsi
; │└
	cmovnsq	%rdx, %rax
	retq
; └
; ┌ @ int.jl:448 within `<invalid>'
	nopw	(%rax,%rax)
; └
```

```julia
julia> @code_native 0xFFFFFFFFFFFFFFFF >> (64 & 63)
	.section	__TEXT,__text,regular,pure_instructions
; ┌ @ int.jl:448 within `>>'
; │┌ @ int.jl:448 within `<='
	xorl	%ecx, %ecx
; │└
; │ @ int.jl:448 within `>>' @ int.jl:442
	cmpq	$63, %rsi
	shrxq	%rsi, %rdi, %rdx
	cmovaq	%rcx, %rdx
; │ @ int.jl:448 within `>>'
; │┌ @ int.jl:51 within `-'
	movq	%rsi, %rax
	negq	%rax
; │└
; │┌ @ int.jl:443 within `<<'
	cmpq	$63, %rax
	shlxq	%rax, %rdi, %rax
	cmovaq	%rcx, %rax
; │└
; │┌ @ int.jl:428 within `<='
	testq	%rsi, %rsi
; │└
	cmovnsq	%rdx, %rax
	retq
; └
; ┌ @ int.jl:448 within `<invalid>'
	nopw	(%rax,%rax)
; └
```

They look the same to me.

jakobnissen: You just need to put it in a function:
```
julia> f(a, b) = >>>(a, b & 63)
f (generic function with 1 method)

julia> code_native(>>>, (Int, Int), debuginfo=:none)
    .section    __TEXT,__text,regular,pure_instructions
    xorl    %ecx, %ecx
    cmpq    $63, %rsi
    shrxq   %rsi, %rdi, %rdx
    cmovaq  %rcx, %rdx
    movq    %rsi, %rax
    negq    %rax
    cmpq    $63, %rax
    shlxq   %rax, %rdi, %rax
    cmovaq  %rcx, %rax
    testq   %rsi, %rsi
    cmovnsq %rdx, %rax
    retq
    nopw    (%rax,%rax)

julia> code_native(f, (Int, Int), debuginfo=:none)
    .section    __TEXT,__text,regular,pure_instructions
    shrxq   %rsi, %rdi, %rax
    retq
    nopw    %cs:(%rax,%rax)
```

jakobnissen: It guess it would be slightly faster still to change `data[i] = x | y << (64 - r)` to `data[i] = x | y << unsigned(64 - r)`

BenJWard: Oh yep that works for me too. So julia is doing some extra stuff to zero a variable, whereas shrxq only cares about the 6 LSBs? I should probably see if I can use this trick in some of my bitpar counting stuff.

jakobnissen: @BenJWard Yes, in Julia's it's equivalent to this like this:
```
function >>>(n::Signed, s::Signed)
    if s < 0
        return n << -s
    elseif s > 63
        return zero(n)
    else
        return LLVM_RIGHT_SHIFT(n, s)
    end
end
```
Using an `unsigned` for `s` removes the first branch, but not the second. Using `& 63` removes both.

PR #92, opened 2020-01-17
Document and improve performance of resize!

jakobnissen:
# Document and improve performance of resize!

Also made `resize!` lazier, so it only actually resizes the underlying data if there is no room. To force resizing underlying data, the user can pass `true` to `copy!`:

```
    resize!(seq, size)

Resize a biological sequence `seq`, to a given `size`. Does not resize the underlying data
array unless the new size does not fit. If `force`, always resize underlying data array.
```

### DIFF
--- a/src/longsequences/transformations.jl
+++ b/src/longsequences/transformations.jl
@@ -3,17 +3,24 @@
 ###
 
 """
-    resize!(seq, size)
+    resize!(seq, size, [force::Bool])
 
-Resize a biological sequence `seq`, to a given `size`.
+Resize a biological sequence `seq`, to a given `size`. Does not resize the underlying data
+array unless the new size does not fit. If `force`, always resize underlying data array.
 """
-function Base.resize!(seq::LongSequence{A}, size::Integer) where {A}
-    if size == length(seq)
-        return seq
-    elseif size < 0
+function Base.resize!(seq::LongSequence{A}, size::Integer, force::Bool=false) where {A}
+    if size < 0
         throw(ArgumentError("size must be non-negative"))
-    else
+    elseif seq.shared
+        # May seem wasteful to orphan here, but if you resize! a shared seq,
+        # you're probably going to write to it anyway.
         return _orphan!(seq, size)
+    else
+        if force | (seq_data_len(A, size) > seq_data_len(A, length(seq)))
+            resize!(seq.data, seq_data_len(A, size))
+        end
+        seq.part = 1:size
+        return seq
     end
 end
 
--- a/src/longsequences/copying.jl
+++ b/src/longsequences/copying.jl
@@ -54,9 +54,26 @@ end
     return dst
 end
 
+"""
+    copy!(seq, src)
+
+In-place copy content of `src` to biological sequence `seq`, resizing `seq` to fit.
+`src` may be any sequence-like object that can be encoded to `seq`.
+
+# Examples
+```
+julia> seq = copy!(dna"TAG", "AACGTM")
+4nt DNA Sequence:
+AACGTM
+
+julia> copy!(seq, [0x61, 0x43, 0x54])
+3nt DNA Sequence:
+ACT
+```
+"""
 function Base.copy!(dst::LongSequence{A}, src::LongSequence{A}) where {A <: Alphabet}
     len = length(src)
-    resize!(dst, len)
+    resize!(dst, len, false)
     @inbounds copyto!(dst, 1, src, 1, len)
 end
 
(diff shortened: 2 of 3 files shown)


jakobnissen:
Needed a few more tests, I think it's good now and ready to be merged.

src/longsequences/transformations.jl
```diff
@@ -3,17 +3,24 @@
...
+function Base.resize!(seq::LongSequence{A}, size::Integer, force::Bool=false) where {A}
+    if size < 0
         throw(ArgumentError("size must be non-negative"))
-    else
+    elseif seq.shared
+        # May seem wasteful to orphan here, but if you resize! a shared seq,
+        # you're probably going to write to it anyway.
         return _orphan!(seq, size)
+    else
+        if force | (seq_data_len(A, size) > seq_data_len(A, length(seq)))
```
BenJWard: Does resize! of a Vector not check is requested size == current size?

jakobnissen: Yes, but it will also shrink the vector if necessary, whereas this piece of code will only shrink resize the data array if the new data doesn't fit (or if forced)

jakobnissen:
@BenJWard Actually, I've been thinking more carefully about this: Do we even want users to be able to `copy!` or `copyto!` from a non-LongSequence to a LongSequence? Or should this *only* be doable using `encode_copy!`? Conceptually, it sort of makes sense that the user should specify whether or not the function should *encode* the input, or just copy it. This is not a breaking change from the current release. If so, the docs here should be changed.

 I think the internal dispatch model for copy!/copyto!/encode_copy! is a little messy (mostly my fault, sorry about that), and it would be nice to streamline it so that there are fewer internal functions, and they always dispatch to the most effective one.

For example, I just discovered `sequence(::FASTA.Record)` doesn't use the most effective dispatch path, and itstead of messing with that, I'd like it to be the case that a developer can simply do `copyto!` or whatever, without thinking too hard about it, and it'll be effective.

Regardless of whether e.g. String -> LongSequence should be able to be done using `copy!`/`copyto!`, I think the method mess can be cleaned up.

BenJWard:
Ok, this is similar to the conundrum I faced about `convert`. For 2.0 I cancelled the ability for the user to `convert(BioSequence, ::String)`. The reason in my mind was that convert is sometimes implicitly called, and the semantics of convert's docs made it clear that conversion should be possible between things that are considered equivalent. 1 and 1.0, and true, and so on.

I decided that whilst all BioSequences may be considered valid strings, not all Strings are valid BioSequences, and trying to convert them may error, and so rather than `convert`, the user must explicitly call `parse`. Maybe we should introduce something like a `parseto!`? 

copyto! for FASTX is fine with me if we are able to agree that FASTA.Records and a BioSequence are equivalent? Maybe a FASTA.Record should even be considered a BioSequence - a BioSequence with some additional metainfo. A new API could be designed for FASTX in which a lot of FASTA and FASTQ records then inherit behaviour from BioSequences.

BenJWard:
I thought sequence(FASTA) used encode_copy for the UInt8 vector case? Which should be fairly quick.

jakobnissen:
I'd argue for `copy!` and `copyto!` for "equivalent" types, e.g. between BioSequences. Then `encode_copy!` for strings, arrays and FASTA records etc. Though of course, not making breaking changes takes priority.

Perhaps I should remove my new, undocumented `copy!` methods where src is a non-BioSequence types, since it hasn't made it into a release yet. Then it's more consistent. Do you agree with this? And then I'll review the dispatch paths to make sure all calls end up where we want them to be.

Yeah, it's fast for `sequence(FASTA)`, but not optimally fast. It dispatches to `encode_copy!(::LongSequence{A}, ::Integer, ::AbstractVector{UInt8}, ::Integer, ::Integer, ::AsciiAlphabet)`, which wastes some time compared to the more efficient `encode_chunks!`. My point here was that it's annoying having to think about where the dispatch ends up, and it should just flow naturally to the optimal place without the developer thinking about how the methods in the `copying.jl` file call each other.

BenJWard:
I think that sounds like a sound proposal, I'm all for it. Would a Vector{DNA} be considered equivalent to LongDNASeq? Because you can argue that the BioSymbols also get encoded just as Strings / Chars and so on do, although the eltype of BioSequences are BioSymbols. I'm fine with either way so long as we are careful to clearly explain to the user what constitutes an equivalent type to a BioSequence.

jakobnissen:
I removed all the `copy!` related code from this PR and moved it into #95 . This PR is much smaller in scope and non-breaking, whereas #95 needs some discussion first.

Issue #86, opened 2020-01-15
Performance review before release


jakobnissen:
So @BenJWard  have mentioned wanting to make a new release of BioSequences that includes performance improvements. We might as well squeeze whatever we have on our performance wishlist into that release.

When #83 and #84 gets merged, I have two smaller PRs lined up which builds on top of those:
* One that improves the performance of printing `LongSequence`. Currently, they are printed base-by-base, which is inefficient in 1.3 and above due to IO operations locking threads. Also, we can exploit the ASCII-related functions I created in the other PRs to make it way faster for ASCII sequences.
* One that introduces a `copy!` method, such that you can in-place modify a single sequence when you are iterating over sequences, similar to what you do for FASTA records. This is just a simple wrapper around `copyto!`, or `encode_copy!` or `encode_chunks`, depending on source and destination types.

### Benchmark
We can use the RC benchmark, created by the Seq guys. This benchmark covers almost all the included benchmarks. I'll be Julia 1.3 on my native computer instead of in a VM. This is comparing all current PRs with commit 61b8faeddb954317a1dbe67f07fd15c592fbd91d. 
```
Before: 100.9 seconds
After:    9.2 seconds
```
## TODO:
* ~~Improve LongSequence printing (PR #91 )~~
* ~~Add copy! method ( PR #87)~~
* ~~Add docs to `copy!` method (PR #95 )~~
* ~~Optimize `length(::LongSequence)` (PR #88 )~~
* ~~Optimize `resize!(::LongSequence)` (PR #92)~~
* ~~Optimize `orphan!(::LongSequence)` (PR #88)~~
* Review bitpar counting kernels ( PR #96 )
* Review ReferenceSequence ( PR #94 )
* ~~SIMD-vectorize reverse complementation ( PR #97 )~~
* ~~Consider optimizing random nucleotide generation (issue #77 )~~
* ~~Improve minhash hashing speed (PR #99)~~

BenJWard:
We should add some docs for the copy! method as well.

jakobnissen:
Good call. I'll add that to #88 perhaps, which has turned into a big pile of small changes anyway. Alternatively, I could just ditch #88 , make one small PR with the `orphan!` fix, that can then be backported to 1.0, and then a big PR with all the optimizations and the `copy!` docs.

One thing I'm not quite sure how to approach: The overhead from incremental writes in `print(io::IO, ::LongSequence)`. This is destroying performance. Here's the options as far as I can see:
[...]

PR #97, opened 2020-01-29
Faster reverse and reverse complement

jakobnissen:
As mentioned in https://github.com/BioJulia/BioSequences.jl/issues/86#issuecomment-579644631.

Improves reversion, complementation and reverse-complement of nucleotides.

This PR needs some more tests, please do not merge it. I will merge it when it has complete coverage.

I think I'm done optimizing this part for now. At 12 µs for a 1 mill nucleotide sequence, I can't possibly think anyone would want it faster.

Edit: Note to self: I should also add more comments in the code.

### DIFF
--- a/src/longsequences/transformations.jl
+++ b/src/longsequences/transformations.jl
@@ -78,6 +78,7 @@ end
 
 _reverse(seq::LongSequence{<:Alphabet}, ::BitsPerSymbol) = reverse!(copy(seq))
 
+# Generic fallback
 function _reverse!(seq::LongSequence{<:Alphabet}, ::BitsPerSymbol)
     i, j = 1, lastindex(seq)
     @inbounds while i < j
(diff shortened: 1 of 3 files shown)


BenJWard:
I'm wondering if the in-place version actually can be compiled to SIMD containing code?

It looks like each iteration is independent:

```julia
@inbounds for i in 1:len >>> 1
        data[i], data[len-i+1] = pred(reversebits(data[len-i+1], B)), pred(reversebits(data[i], B))
    end
```
The first iter will touch data[1] & data [len-1+2], and AFAICT no other iteration will touch those two elements, the second iteration will touch second data[2] & data[len-2+1] and so on.

Perhaps `@simd ivdep` will persuade the julia compiler to vectorize?

jakobnissen:
> Perhaps `@simd ivdep` will persuade the julia compiler to vectorize?

Haha amazing, I didn't even know about `@simd ivdep`! 😄  Yes, it makes it vectorize, and significantly faster. On my computer, it now does `reverse_complement!` on 100k 2-but nucs in 1.14 seconds. I've still kept all the methods, since non-inplace `reverse!` still has the advantage that it copies the sequence, reverses and complements it in one pass.

Still needs tests.

jakobnissen:
Will merge this ASAP, unless codecov/CI objects.

PR #101, opened 2020-05-29
Support LTS

CiaranOMara:
The `isnothing` function doesn't exist in julia v1.0.


kescobo:
LGTM

kescobo:
Maybe worth using travis for testing on Windows rather than appveyor now that that's supported

CiaranOMara:
Last time I tried Windows on Travis it was inconsistent (Cygwin stuff). Though it is probably more consistent these days, anyhow, GitHub actions will probably be the next thing when there is greater clarity around its configuration. Also, there's the existing configuration of this repository... I'm inclined to leave Appveyor as it is for the moment.

BenJWard:
I think BioSymbols might need to be made to support LTS before travis starts passing.

jgreener64:
I'm appreciating the effort going in here to support LTS.

BenJWard:
I may not have newer packages support less than 1.5 involved in the genome assembler if I rely on the 1.5 struct with mutable refs optimisation for performance. But most packages should if they can.

PR #106, opened 2020-06-04
Use stable RNG (take two)

### LINKED ISSUE #77: Faster random LongSequence{NucleicAcidAlphabet{4}} generation
I was goofing around the other day and discovered a way to make generation of `LongSequence{NucleicAcidAlphabet{4}}` faster. It's probably fast enough already for any practical application, but in the words of Jeff Bezanson: "[MOAR PERFORMANCE](https://www.youtube.com/watch?v=TPuJsgyu87U&t=338s)". And also, this new code is shorter.
I'm aware that BioSequences guarantees reproducible random sequence generation, and a new algorithm messes with that, so I make this issue here instead of a PR, so we can perhaps use it in a future version of BioSequences.

The new function looks like this:
```
function randseq2(rng::AbstractRNG, A::NucleicAcidAlphabet{4}, len::Integer)
    seq = LongSequence{typeof(A)}(len)
    data = seq.data
    rand!(rng, data)
    @inbounds for i in eachindex(data)
        nuc = 0x1111111111111111
        mask = data[i]
        nuc = ((nuc & mask) << 1) | (nuc & ~mask)
        mask >>>= 1
        nuc = ((nuc & mask) << 2) | (nuc & ~mask)
        data[i] = nuc
    end
    if len > 0
        @inbounds data[end] >>>= ((16 - (len & 15)) & 15) << 2
    end
    return seq
end
```
And it's fast because it uses `rand!` to generate all the randomness at once, and because it SIMD-vectorizes. It's about 8x faster for long sequences.

jakobnissen:
Resolves #77 (take two, I messed up my branches on the other one)

This uses StableRNGs. With its simple implementation, stable RNGs are slower than Julia's default, meaning that this PR slightly slows down random seq generation. However, I think the slowdown and additional dependency is worth it to guarantee reproducible random sequences across major BioJulia versions.

@kescobo I still don't know how to properly add dependencies to Julia projects. Could you please check if the `Project.toml` is correct? In particular, do I need to add something to the compat part?

For comparison, here are some simple benchmarks when creating sequences of length 10,000:
```
                        before ns    after ns
NucleicAcidAlphabet{2}   521           390
NucleicAcidAlphabet{4}  7440           873
AminoAcidAlphabet      62700         92114
SamplerUniform         61440         62700
SamplerWeighted        79000         83000
```

So yeah, the effect is pretty minor, and it's even faster for random nucleotide generation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,6 +10,7 @@ Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
 IndexableBitVectors = "1cb3b9ac-1ffd-5777-9e6b-a3d42300664d"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Twiddle = "7200193e-83a8-5a55-b20d-5d36d44a0795"
 
 [compat]
--- a/src/BioSequences.jl
+++ b/src/BioSequences.jl
@@ -266,6 +266,7 @@ import Twiddle: enumerate_nibbles,
     count_nonzero_bitpairs,
     repeatpattern
 using Random
+using StableRNGs
 
 BioSymbols.gap(::Type{Char}) = '-'
 
--- a/src/longsequences/randseq.jl
+++ b/src/longsequences/randseq.jl
@@ -7,7 +7,8 @@
 ### This file is a part of BioJulia.
 ### License is MIT: https://github.com/BioJulia/BioSequences.jl/blob/master/LICENSE.md
 
-import Random: Sampler, GLOBAL_RNG
+import Random: Sampler, rand!
+const DEFAULT_RNG = StableRNG(rand(Int) & typemax(Int))
 
 """
     SamplerUniform{T}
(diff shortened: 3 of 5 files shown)


BenJWard:
Nice! I think given actually a lot of random seq generation is basically only ever used in simulations and testing, that this is a slowdown that is acceptable if it guarantees correctness and stability - a bit like the Automa tradeoff again, a small slowdown is fine if it buys you confidence and robustness.

kescobo:
@jakobnissen Yes, I think there needs to be a `[compat]` entry for all packages that aren't part of the Stdlib to get merged into the General registry. Looks like there's only 0.1.0 and 0.1.1, so should be as simple as `StableRNGs = "0.1".

KristofferC:
I don't understand the `rand(Int) & typemax(Int)` seed for the `StableRNG`. This will happen during precompilation time and be "frozen" at that point. Did you mean to put this seeding in `__init__`? Or what is the purpose of it?

jakobnissen:
@KristofferC Yeah, good catch. I'm pretty sure I just didn't understand how precompilation worked when I wrote this. Now I'm wondering if there even is a point to having `StableRNGs` by default in `BioSequences`. Presumably, if the end user cares about reproducible RNG, they can just add it themselves.

KristofferC:
Many functions here don't take an RNG object so I'm not sure how you would get reproducibility anyway. Feels like the simplest would be to have these functions accept an RNG but default to the Base default one and indeed allow others to pass their own stable rng.

kescobo:
#187

PR #107, opened 2020-06-04
Add StableRNGs Compat entry

kescobo:
ref https://github.com/BioJulia/BioSequences.jl/pull/106#issuecomment-638985043

cc @CiaranOMara

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,6 +18,7 @@ BioGenerics = "0.1.0"
 BioSymbols = "4.0.0"
 Combinatorics = "1.0.0"
 IndexableBitVectors = "1.0.0"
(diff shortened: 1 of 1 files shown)


CiaranOMara:
That entry looks good to me. It'll cover the range [0.1, 0.2).
 
No idea how vital that bug fix was. But, Pkg will choose the highest version anyway.
- https://github.com/rfourquet/StableRNGs.jl/compare/v0.1.0...v0.1.1

kescobo:
@CiaranOMara Right - presumably, if that package author is following Semver, there shouldn't be any reason for someone to use 0.1.0 when 0.1.1 is available.

PR #108, opened 2020-06-05
Fixup addition with a BitIndex

BenJWard:
## Types of changes

* [ ] :sparkles: New feature (A non-breaking change which adds functionality).
* [X] :bug: Bug fix (A non-breaking change, which fixes an issue).

There are cases that can occur in `copyto!` methods where depending on how the arithmetic to compute increments of BitIndex go - if they return a UInt and try to do the addition with the BitIndex, then there is a method error. This is because addition of a number to a BitIndex is only defined for `BitIndex + Int`, but really to make sure stuff like this does not crop up I've edited the + and - methods to work for `BitIndex & Integer`, which permits both signed and unsigned integers.


BenJWard:
I stumbled across this with the `copyto!` method for FASTX, the one which requires all the offsets and N. I think ReadDatastores uses the method but must use a UInt for the N value which trips this.

Issue #118, opened 2020-07-21
Future of subsequences


jakobnissen:
Pinging @BenJWard @CiaranOMara perhaps this discussion should be part of #102 , but I do feel like the scope is a little broader here.

## The situation

We currently are working on multiple different "subsequence" like types:
* Ben is (has been?) working on NTupleKmers
* I have a working prototype of a sequence view, similar to an array view (link coming soon)
* I'm playing around with 256-bit kmers

This comes on top of the already pretty large subsequence codebase:
* LongSequence's copy-on-write (COW) mechanism that allows a LongSequence to actually be a subsequence
* The recently revamped Mer types.

So I'm concerned if not done right, we will end up with an enormous unelegant codebase that might still leave some bases uncovered. So: What do we really need? As I see it, we really just need:

* An arbitrary-length subsequence, preferrably allocation-free.
* An efficient type for kmers. Ideally, this would be flexible in size as well, but in my experience, kmers with k > 127 are rarely used.

## My proposal
In this issue, I make two arguments:
 
__Argument one: We should move completely to either NTupleKmers if they work, else N-bit kmers__
While a maximum of K = 32 is good enough for many purposes, the usage of longer kmers is common enough that it's an awkward ceiling. By leveraging BitIntegers.jl, we could change the Kmer code to wrap an integer up to 1024-bit integers. Essentially its structure would be:
```
struct Kmer{K, A <: Alphabet, T <: Unsigned}
    data::T
end
```

For longer kmers like 51-mers (common in metagenomics), 127-mers (sometimes used in assembly), and perhaps even higher values than that, a truly variable-length kmer type like NTupleKmers is obviously best. Furthermore, I personally find it likely that NTuples will be even further optimized by the Julia core devs in the future since they are integral to immutable arrays.

However, NTupleKmers may be very difficult to write efficient code for. If that is the case, we can simplify the future codebase a great deal by having our kmers simply wrap a BitInteger type which is already aggressively optimized by LLVM. I am working on a proof of concept for this for show. Unfortunately, it adds another dependency for BioSequences.

* __Proposal 1__: Remove our current Kmer types. If NTupleKmers work well, use those solely. Otherwise, add BitIntegers.jl as a dependency, and make a kmer type that parameterizes the integer type that stores its data

__Argument two: We should remove the COW mechanism of LongSequence__
[...]

PR #119, opened 2020-07-25
Remove LongSequence copy-on-write

jakobnissen:
As discussed in #118 , this PR removes the entire copy-on-write mechanism for `LongSequence`. As a result, the code is now a little cleaner and easier to work with.

### Notable changes:
* `LongSequence` now has a new layout:
Before:
```
mutable struct LongSequence{A <: Alphabet}
    data::Vector{UInt64}
    part::UnitRange{Int}
    shared::Bool
end
```
After:
```
mutable struct LongSequence{A <: Alphabet}
    data::Vector{UInt64}
    len::Int
end
```
* Slicing a `LongSequence` now copies the underlying data - not all of it, just the relevant part.
* A few function has been made more efficient due to lack of `orphan!` calls. Gains are very modest tbh.
* Refactored the `LongSequence` hashing code to use functions instead of macros (more generic, to be used for SeqView eventually). Behaviour and performance is identical.

### What needs to be done
I would like to take another pass over the codebase and remove any unused functions and simplify what can be simplified. I'm a little disappointed that not more lines were deleted.

Also now might be a good time to do some spring cleaning of the tests. These can presumably be simplified a lot since we no longer need to test orphan! or subsequences. Also, the code coverage could be better.

### DIFF
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -44,43 +44,11 @@ DNA_A
 !!! note
     Some types such as `Kmer` can be indexed using integers but not using ranges.
 
-For `LongSequence` types, indexing a sequence by range creates a subsequence
-of the original sequence.
-Unlike `Arrays` in the standard library, creating this subsequence is copy-free:
-the subsequence simply points to the original `LongSequence` data with its range.
-You may think that this is unsafe because modifying subsequences propagates to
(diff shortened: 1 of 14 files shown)


jakobnissen:
OK, this is good to go. No rush merging this though, maybe let's have it laying for a bit until I get a SeqView PR up

Issue #102, opened 2020-05-29
julia 1.5: Time for views of a sequence?


BenJWard:
@jakobnissen @CiaranOMara 

So julia 1.5 now means structs that have references to mutable are now stack-allocated. 
I'm very excited about this. As you know, every `LongSequence` are mutable and every `LongSequence` owns its data.

Yes, there is a copy on write mechanism, but that's just an implementation detail to remove copying unless absolutely nessecery.

If you create many views over a sequence - taking advantage of the mechanism, you avoid copying, but you still need to allocate many LongSequence objects on the heap and trigger the gc and all that.

I've long wanted to make proper allocation free views of a sequence - a true view of fixed start and end, where if you edit a base you DO indeed edit the base in the underlying sequence (contrary to the copy on write mechanism which mean the seqs still owns its own data - even if sharing a vec temporarily for performance reasons). But there was never any point as they would just allocate to. Perhaps 1.5 is the time to try something - what do you guys think? I might be a massive boon, for example making kmers with K > 64?

jakobnissen:
Yes, se could make a SeqView object - or whatever we should call it.

I'm not sure what the use case is.When do you create so many views that the microsecond-delay caused by an allocation matters? Maybe for genome assembly purposes, where K=127 is sometimes (rarely) used. Even then, a strict wrapping four 64-bit integers may be way more efficient. Do you have a use case, maybe for your GenomeGraphs?

Perhaps it doesn't matter that I don't have a use case, if you build the tool, sometimes people will figure out what to do with it. If we do it, I think we should make it unsafe and high performance: Just wrap the source array and a UnitRange in a struct. If people then resize the underlying array, the segfault is on them.

BenJWard:
There's also all the indirections and lack of contiguous memory layout if put in a vector in addition to the creation.

So in the group's genome assembly research, we've got an algorithm that unlike any other assembler (even the reportedly reliable ones), really does give you reliable repeat expansion of repetitive regions between two haplotype-specific anchors, no matter the ploidy. It turns out many assemblers have diploid bias and screw up even simple triploid scenarios generated in Pseudoseq where we know the answer, because even though they are not aiming for a diploid assembler or phasing or whatever, they have human genomes on the brain if you like, so when heuristics are dreamed up, there are edge cases and undocumented assumptions that crop up. 

[...]

PR #120, opened 2020-07-26
Add sequence view

jakobnissen:
This adds a sequence view `SeqView` as discussed in #102 .

Because a `LongSequence` and `SeqView` share the same underlying encoding, many of the old `LongSequence` methods has been changed to now take a `SeqOrView` (which is just `Union{LongSequence, SeqView}`)

You can construct a view with
* `SeqView(s::LongSequence, ::UnitRange)`
* `@view my_seq[1:5]`
* `SeqView(::Vector{UInt64}, ::UnitRange)` (this one doesn't boundscheck)

Mutating methods that change the length of the view are intentionally not implemented, such as
* `filter!`
* `map!`
* `resize!`
* `ungap!`


BenJWard:
Is editing a `LongSequence`, making it smaller say as a result of `ungap!` or `resize!` going to present a problem for a previously created view of say `@view seq[90:100]` if the sequence is resized to only 90bp?

src/longsequences/transformations.jl
```diff
@@ -11,50 +11,40 @@ array unless the new size does not fit. If `force`, always resize underlying dat
...
-            if index(next) != j
-                seq.data[j] = datum
-                datum = 0
-                j = index(next)
+function Base.filter!(f, seq::LongSequence)
+    writeindex = bitindex(seq, 0)
+    chunk = zero(UInt64)
+    bps = bits_per_symbol(seq)
+    for i in eachindex(seq)
+        encoded_symbol = extract_encoded_element(bitindex(seq, i), encoded_data(seq))
```
BenJWard, line 28: Doesn't inbounds_getindex do these two lines already?

jakobnissen: Yep, but I'm using the encoded symbol a few lines down, and don't want to redo the calculation,

jakobnissen:
> Is editing a `LongSequence`, making it smaller say as a result of `ungap!` or `resize!` going to present a problem for a previously created view of say `@view seq[90:100]` if the sequence is resized to only 90bp?

Yes, the view is completely unsafe. In theory it could even segfault if the underlying array is truncated. Perhaps I could add another boundscheck when accessing a view that the underlying buffer must be long enough? That would slow down boundschecked access just a little.

BenJWard:
Ok, I'm happy with whatever behaviour closest resembles what julians already expect from a view, since that will be most consistent. We'll just make sure to document any pitfalls properly.

CiaranOMara:
I was proposing "Sub" as a prefix convention.

jakobnissen:
I think this has been hanging around long enough - I'm merging this now, and will begin using `master` branch as my daily driver. Hopefully I can hammer out any remaining bugs that way.

PR #123, opened 2020-09-29
Make same-type conversion noop

jakobnissen:
On master, the following method is defined:
```
function Base.convert(::Type{T}, seq::LongSequence{<:NucleicAcidAlphabet}) where
         {T<:LongSequence{<:NucleicAcidAlphabet}}
    return T(seq)
end
```
Seems reasonable enough, right? We want to be able to convert between nucleotide sequences.
Except that it causes this:
```
julia> seq = dna"TAG"
3nt DNA Sequence:
TAG

julia> convert(typeof(seq), seq) === seq
false
```
What's the big deal? If we call convert, we probably want a separate copy anyway.
Well, the problem is that Julia sometimes implicitly calls convert:
```
julia> struct Foo{T}
           x::T
       end

julia> seq = dna"TAG"
3nt DNA Sequence:
TAG

julia> foo = Foo{typeof(seq)}(seq)
Foo{LongSequence{DNAAlphabet{4}}}(TAG)

julia> foo.x === seq
false
```
As far as I can tell, that makes it impossible to wrap a `LongSequence`! Clearly, that's completely unworkable.
This PR simply defines `Base.convert(::Type{T}, seq::T) where {T <: LongSequence} = seq`.
Note that this may be considered breaking behaviour, but I think this needs to happen.

### DIFF
--- a/src/longsequences/conversion.jl
+++ b/src/longsequences/conversion.jl
@@ -19,6 +19,8 @@ end
 ###
 ### Conversion
 ###
+Base.convert(::Type{T}, seq::T) where {T <: LongSequence} = seq
+Base.convert(::Type{T}, seq::T) where {T <: LongSequence{<:NucleicAcidAlphabet}} = seq
 
 function Base.convert(::Type{T}, seq::LongSequence{<:NucleicAcidAlphabet}) where
          {T<:LongSequence{<:NucleicAcidAlphabet}}
--- a/test/longsequences/conversion.jl
+++ b/test/longsequences/conversion.jl
@@ -121,6 +121,17 @@ end
     end
 end
 
+@testset "Convert to same type" begin
+	function test_same_conversion(seq)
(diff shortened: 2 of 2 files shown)


jakobnissen:
cc @BenJWard just need your blessings (because it's arguably breaking), then I'll merge this.

BenJWard:
Is there a convention already set by other mutable types (say vectors) for this conversion behaviour? In principle I'm totally fine with it.

BenJWard:
The current convention for Array like objects is `convert(::Type{T}, a::AbstractArray) where {T<:Array} = a isa T ? a : T(a)`, so I'm trilled if this goes in and fixes some odd edge cases.

jakobnissen:
Yes, this is documented in the Julia docs: https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Mutable-collections

PR #128, opened 2021-01-14
Fix reversing seq with unused data

jakobnissen:
MWE to reproduce bug:
```
julia> using BioSequences

julia> seq = ungap!(dna"TAGC------------ACC")
7nt DNA Sequence:
TAGCACC

julia> reverse_complement(seq)
7nt DNA Sequence:
----GGT
```
What happened was that the reversing functions assumed that `length(seq.data) == seq_data_len(typeof(Alphabet(seq)), length(seq)`, which is not a valid guarantee (i.e: `LongSequences` are allowed to have noncoding elements at the end of their `data` vector).

### DIFF
--- a/src/longsequences/transformations.jl
+++ b/src/longsequences/transformations.jl
@@ -69,7 +69,7 @@ Base.reverse(seq::LongSequence{<:Alphabet}) = _reverse(seq, BitsPerSymbol(seq))
 @inline function _reverse(seq::LongSequence{A}, B::BT) where {A <: Alphabet,
     BT <: Union{BitsPerSymbol{2}, BitsPerSymbol{4}, BitsPerSymbol{8}}}
     cp = LongSequence{A}(unsigned(length(seq)))
-    reverse_data_copy!(identity, cp.data, seq.data, B)
+    reverse_data_copy!(identity, cp.data, seq.data, seq_data_len(seq) % UInt, B)
     return zero_offset!(cp)
 end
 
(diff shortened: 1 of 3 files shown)


src/longsequences/transformations.jl
```diff
@@ -69,7 +69,7 @@ Base.reverse(seq::LongSequence{<:Alphabet}) = _reverse(seq, BitsPerSymbol(seq))
 @inline function _reverse(seq::LongSequence{A}, B::BT) where {A <: Alphabet,
     BT <: Union{BitsPerSymbol{2}, BitsPerSymbol{4}, BitsPerSymbol{8}}}
     cp = LongSequence{A}(unsigned(length(seq)))
-    reverse_data_copy!(identity, cp.data, seq.data, B)
+    reverse_data_copy!(identity, cp.data, seq.data, seq_data_len(seq) % UInt, B)
```
cjprybol, line 72: Taking the remainder like this is a clever way of converting an integer to an unsigned integer. Didn't realize what was going on here until I tried it

```julia
julia> x = 1
1

julia> UInt(x)
0x0000000000000001

julia> x % UInt
0x0000000000000001

julia> UInt(x) == (x % UInt)
true

julia> BenchmarkTools.@benchmark UInt($(rand(Int)))
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     0.059 ns (0.00% GC)
  median time:      0.070 ns (0.00% GC)
  mean time:        0.074 ns (0.00% GC)
  maximum time:     0.539 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> BenchmarkTools.@benchmark $(rand(Int)) % UInt
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     0.040 ns (0.00% GC)
  median time:      0.040 ns (0.00% GC)
  mean time:        0.045 ns (0.00% GC)
  maximum time:     15.710 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```
```

PR #129, opened 2021-01-15
Removed minhash

jakobnissen:
This PR removes the minhash functionality of BioSequences.

The reason is that I think minhashing is important enough to warrant its own package, while also being niche enough that normal users of BioSequences.jl should not have to load all the extra code.

Putting more effort into fine-tuning the minhash implementation does pay off: Here is a benchmark of the current BioSequences.jl implementation versus my package Kash.jl:
```
using BioSequences
using FASTX
using Kash

# Make a function which imitates the BioSequence's interface for minhashing vectors
function sketch_vector(v)
    sketcher = KmerSketcher{DNAMer{16}}(10000)
    foreach(s ->update!(sketcher, s), v)
    KmerHashes(sketcher)
end

# BioSequences.jl
seqs = [randdnaseq(rand(15000:25000)) for i in 1:100]
sketch = minhash(seqs, 16, 10000)
sketch = open(io -> minhash(io, 16, 10000), FASTA.Reader, "/Users/jakobnissen/Downloads/fasta.fna")

# Kash.jl
sketch = sketch_vector(seqs)
sketch = open(io -> kmer_minhash(io, 10000, Val(16)), "/Users/jakobnissen/Downloads/fasta.fna")
```
The file has 50,000 sequences each of length 2,000. Results (median execution time):
```
BioSequences, vector: 53 ms
BioSequences, file: 3141 ms
Kash, vector:         16 ms (3.3x)
Kash, file:          630 ms (5.0x)
```


kescobo:
I was the one that originally put it in, back when Bio.jl was a giant monorepo. At one point there was discussion of putting it in a `Kmers.jl` or something.

All of which is to say I agree with this. Not sure what the release schedule is like, but we may want to do one release with a deprecation warning first, no?

jakobnissen:
I mean, there is no rush to get this removed. But maybe we can add a deprecation warning in the next minor release, then remove it for the next major? That sounds like a good idea.

BenJWard:
I'm totally fine with that. I think this year, with new kmers, several bug fixes, interface tweaks, and the views will be a year of a major v3 shakeup.

jakobnissen:
Closing this PR, as minhashing has been removed on the v3 branch, and the deprecation has been added to the master branch

PR #132, opened 2021-02-25
Deprecate minhash

jakobnissen:
See pr #129 .


jakobnissen:
This is just codecov being weird - I'm merging.

kescobo:
Just to be sure, there's going to be a minor release with this deprecation included before it's removed entirely, right?

jakobnissen:
Sure. We could ship that minor release soon, perhaps. I'm thinking v3 should land August, at latest.

PR #154, opened 2021-04-21
Add fast path for sequence equality

jakobnissen:
This adds a fast path to default sequence equality comparison so it checks for egal first. No behaviour is changed.


BenJWard:
This has an issue with `join`? Otherwise LGTM.

BenJWard (approved):

Issue #141, opened 2021-03-11
Streamline find functions to behave like Base's


jakobnissen:
Right now, you can do:
```
julia> findfirst(DNA_A, dna"TAGCGATGA")
2
```
Whereas it should be:
```
julia> findfirst(isequal(DNA_A), dna"TAGCGATGA")
2
```
The latter example currently errors. This happens because the predicate version of find* in BioSequences are currently restricted to subtypes of Function. But there are many function-like objects (such as e.g. Base.Fix2), which are not instances of Function.

We should remove `findfirst(DNA_A, dna"TAG")`, and make sure these functions behave as expected.

kescobo:
A counterpoint - it depend on whether we think of `BioSequence`s more like strings, or more like generic iterables, since:

```
julia> findfirst('a', "garden")
2

julia> findfirst(3, 1:5)
ERROR: MethodError: no method matching findfirst(::Int64, ::UnitRange{Int64})
# ...

julia> findfirst("ard", "garden")
2:4
```

I think it might make sense to keep this out of the next major release, and then we can always add it back later without it being a breaking change

jakobnissen:
Another option, perhaps a better compromise:
Currently `findfirst(dna"A", dna"TNG")` and `findfirst(DNA_A, dna"TNG")` behave differently. The former will report the `N` as being an `A`, because they are compatible, the latter will not.

Perhaps we can remove that consistency by instead having them both behave in the former manner? For exact sequence comparison, one can then use `findfirst(isequal(DNA_A), dna"TNG")`.

I would still like to restrict the latter to only be able to search for instances of `BioSymbol`.

kescobo:
Oh, that's interesting. I don't have a good sense of what makes sense as an API. 

Are you saying that you do not want to allow searches akin to `findfirst("ard", "garden")`? Given the complexity of sequence comparison in biology, I think that's perfectly justifiable. I can imagine a function that takes a scoring matrix and finds the first match, but then it's not really `findfirst` anymore.

jakobnissen:
No, in the merged fork where I deleted the regular findfirst, you can still do `findfirst("ard", "garden")`, and it does this specific search where it searches for the first compatible subsequence given ambiguous nucleotides.

It's the strange situation with searching for single symbols in sequences I want to change. Either:
* Remove `findfirst('a', "garden")` and only rely on `findfirst(isequal('a'), "garden")`, or
[...]

PR #166, opened 2021-06-24
Add findfirst method for BioSymbol

jakobnissen:
See #141

This PR adds `findfirst` and `findlast` methods of the type `findfirst(::BioSymbol, ::BioSequence)`. How it currently works is:
* `findfirst(::BioSymbol, ::BioSequence)` allows for compatible but different symbols, i.e. `dna"TACNA"` has a potential `G` at position 4. This is the new thing in the PR.
* `findfirst(::BioSequence, ::BioSequence)` likewise allows compatible symbols, just like before.
* To get literal match against a symbol, do `findfirst(isequal(DNA_A), dna"GGA")`
* There is currently no way to search for literal matches between two biological sequences. Perhaps there should be, but I'm not sure what the syntax should be.

Issues / things to mull over:
* Currently, search uses `compatbits` for the element `BioSymbol` type, and this is not generically defined for user-defined biosymbols (see https://github.com/BioJulia/BioSymbols.jl/issues/42).
* There is no way to search for literal biosequence matches.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ authors = ["Ben J. Ward <benjward@protonmail.com>"]
 version = "3.0.0"
 
 [deps]
-Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 BioSymbols = "3c28c6f8-a34d-59c4-9654-267d177fcfa9"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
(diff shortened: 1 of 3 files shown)


BenJWard:
I think this is fine.

As for your question as to a non-compatibility version of `findfirst(::BioSequence, ::BioSequence)`, in the future.

Have you considered having the method accept an optional keyword that toggles the behaviour between finding the first compatible subsequence, and the first exact subsequence? Maybe say if `strict = true` then the symbols subsequence must be exact, not just compatible.

jakobnissen:
Yes, that sounds like a good idea. I'll see if I can make a PR.

PR #176, opened 2021-07-24
Approximate search now returns nothing when no match is found

jakobnissen:
Approximate search functions now return `nothing` when no match is found. These are:
* `approxsearch`
* `approxrsearch`
* `approxsearchindex`
* `approxrsearchindex`

Behaviour is otherwise unaltered. Specifically, search for an empty sequence is valid, and returns an empty range, just like before. Weird, but it's consistent with how string search works in Base.

### DIFF
--- a/docs/src/sequence_search.md
+++ b/docs/src/sequence_search.md
@@ -89,7 +89,6 @@ sequence:
 julia> seq = dna"ACAGCGTAGCT";
 
 julia> approxsearch(seq, dna"AGGG", 0)  # nothing matches with no errors
-0:-1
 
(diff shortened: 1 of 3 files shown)


jakobnissen:
Okay, this is good to go. Leaving it up for some days or weeks if anyone wants to review it.

CiaranOMara:
Is it intended that the `isempty` checks return a range?

I haven't thought this through, though; I wonder whether a range result could be useful for chaining/currying searches in that the result provides the start position for the next search, removing the need to track position?

```julia
findfirst(seq, vague_query) |> (r) -> findfirst(seq, query, first(r), last(r))  |> (r) -> findfirst(seq, specific_query, first(r), last(r))
```

PR #182, opened 2021-08-06
Make macro literals dynamic

jakobnissen:
Currently, as documented in the documentation on "construction", string literals return a statically allocated `LongSequence`. This is great for performance, but can lead to unexpected bugs:

```julia
julia> using BioSequences

julia> start_codon() = dna"TAG"
start_codon (generic function with 1 method)

julia> push!(start_codon(), DNA_G)
4nt DNA Sequence:
TAGG

julia> push!(start_codon(), DNA_G)
5nt DNA Sequence:
TAGGG

julia> push!(start_codon(), DNA_G)
6nt DNA Sequence:
TAGGGG
```

This PR makes it allocate a new sequence at every call. To get the old behaviour back, one can use the "static" flag: `dna"TAG"s`. In other words, the old behaviour was fast by default, opt-in safe, this PR makes it safe by default, opt-in fast.

I'm not 100% sure it's a good change, but like 80% sure. I'll leave the PR up for a week or two, and merge it if no-one has any complaints.

### DIFF
--- a/docs/src/construction.md
+++ b/docs/src/construction.md
@@ -165,144 +165,41 @@ julia> aa"ARNDCQEGHILKMFPSTWYVX"
 ARNDCQEGHILKMFPSTWYVX
 ```
 
-However, it should be noted that by default these sequence literals
-allocate the `LongSequence` object before the code containing the sequence
-literal is run.
-This means there may be occasions where your program does not behave as you
-first expect.
-For example consider the following code:
+Unlike most string literal macros in Julia, the resulting sequences are allocated
+at runtime, not at compile time - so called _dynamic allocation_. Most other Julia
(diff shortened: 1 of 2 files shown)


jakobnissen:
The counter-argument to this PR is that it can be _very_ confusing for people well-versed in Julia if `@dna_str` suddenly begins behaving differently from all other string literal macros. I'm ambivalent here, so I will keep this open until someone else chimes in.

BenJWard:
I thought that the macros did already do this. That they were static by default, but you could make them dynamic with a flag?

jakobnissen:
They do. This PR is about making them dynamic by default. But I'm getting more convinced it's a bad idea and they should just keep being static as they are.

Issue #187, opened 2021-10-05
Stop shipping with StableRNGs?


jakobnissen:
Right now, BioSequences ship with StableRNGs for reproducible RNG behaviour. But... why, exactly? Do most users really care about precise RNG reproducibility?

And if they do, they would need to manually set the seed of an RNG object anyway, and then they might as well just use StableRNGs themselves and pass in the RNG.

Should we just remove this dependency for 3.0?

kescobo:
Yeah - I don't think we need it. I think sometimes, it's nice to have as a test dependency if you want to have reproducible random tests generated (though lately, I've been just building examples by hand).

jakobnissen:
Agreed, we should keep it as a test dependency.

BenJWard:
Yeah, let's get rid of shipping it w/ BioSequences!

PR #188, opened 2021-10-05
Remove dependency on StableRNGs

jakobnissen:
Remove StableRNGs from dependencies.
Methods for LongSequence generation now default to using Random.DEFAULT_RNG.
StableRNGs is kept as a test dependency to test reproducibility.
See issue #187.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,7 +6,6 @@ version = "3.0.0"
 [deps]
 BioSymbols = "3c28c6f8-a34d-59c4-9654-267d177fcfa9"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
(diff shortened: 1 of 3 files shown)


BenJWard (approved):

KristofferC:
>  StableRNGs is kept as a test dependency to test reproducibility.

It doesn't look it StableRNG is used for the tests?

Here is the default global rng used:

https://github.com/BioJulia/BioSequences.jl/blob/0ba3d4b2e1dab04b77e244141501e6b33d53b975/test/runtests.jl#L38

and here:

https://github.com/BioJulia/BioSequences.jl/blob/0ba3d4b2e1dab04b77e244141501e6b33d53b975/test/runtests.jl#L60

etc

PR #196, opened 2021-10-22
Remove sequence conversion to/from String and Vector

jakobnissen:
# Remove the ability to convert BioSequence to/rom String and Vector

This is a speculative, breaking change for v3. I'm pretty sure it's the right way to go, but it needs discussion.
Currently, we allow conversion between `BioSequence` and `String` and between `BioSequence` and `Vector`. I think having constructors between these are reasonable, but not conversion.

What's the difference between conversion and construction anyway? The Base docs on this issue states:

> since convert can be called implicitly, its methods are restricted to cases that are considered "safe" or "unsurprising". convert will only convert between types that represent the same basic kind of thing (e.g. different representations of numbers, or different string encodings).

But what exactly is "safe" and "unsurprising" is in the eye of the beholder.
To me, it makes sense to `convert` between values that are considered equal. For example, `2.0` can be converted to `0x02` because `2.0 == 0x02`.

However, since
```julia
julia> [DNA_A, DNA_G] == dna"AG"
false

julia> "AG" == dna"AG"
false
```

I think we should not be allowing conversion here.

Furthermore, `convert` is called implicitly. This is often acceptable for small bitstypes like integers where performance is not a concern. But implicitly constructing long strings on the heap is not great. Especially because the implications of implicitly creating copies of mutable structs are complicated.

- [X] :blue_book: I have updated and added relevant docstrings, in a manner consistent with the [documentation styleguide](https://docs.julialang.org/en/v1/manual/documentation/).
- [X] :blue_book: I have added or updated relevant user and developer manuals/documentation in `docs/src/`.
- [X] :ok: There are unit tests that cover the code changes I have made.
- [X] :ok: The unit tests cover my code changes AND they pass.
- [X] :pencil: I have added an entry to the `[UNRELEASED]` section of the manually curated `CHANGELOG.md` file for this repository.
- [X] :ok: All changes should be compatible with the latest stable version of Julia.
- [X] :thought_balloon: I have commented liberally for any complex pieces of internal code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   `LongRNASeq` -> `LongRNA` and `LongDNASeq` -> `LongDNA`
 - The interface for `Alphabet` and `BioSequence` is now more clearly defined, documented, and tested.
 - The constructor `LongSequence{A}(::Integer)` has been removed in favor of `LongSequence{A}(undef, ::Integer)`.
+- Biological sequences can no longer be converted to/from strings and vectors.
 
 ## [2.0.1]
 ### Changed
--- a/docs/src/construction.md
+++ b/docs/src/construction.md
@@ -112,34 +112,23 @@ true
 
 ## Conversion of sequence types
 
-You can convert between sequence types.
+You can convert between sequence types, if the sequences are compatible - that is, if the source sequence does not contain symbols that are un-encodable by the destination type.
 ```jldoctest
-julia> dna = dna"TTANGTAGACCG"
+julia> dna = dna"TTACGTAGACCG"
 12nt DNA Sequence:
-TTANGTAGACCG
+TTACGTAGACCG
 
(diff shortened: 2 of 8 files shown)


CiaranOMara:
The implicit conversion only occurs when assigning to declared types ([When-is-convert-called?](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#When-is-convert-called?)). An end user would need to go out of their way to set up a scenario for implicit conversion to occur. So I'm not worried about large amounts of data being unintentionally converted.

In terms of the `Vector` case presented, I'd argue that the elements of the `Vector` and `BioSequence` containters represent symbols, and if those elements represent the same symbols in the same order, then at very least `[DNA_A, DNA_G] == collect(dna"AG")` should be `true`.

I think the `String` case is similar and that `collect("AG") == collect(Char, dna"AG") ` yielding `true` is reasonable.

CiaranOMara:
> I think we can already construct sequences from iterators

I see you added that in #161.

jakobnissen:
Okay, merging this as soon at tests pass

PR #200, opened 2021-10-26
Rename stringbyte and avoid pirating BioSymbols

### LINKED ISSUE #198: Don't recommend users engage in type piracy with `stringbyte`
So, some time ago I added the methods `stringbyte(::Alphabet, ::UInt8)` and `stringbyte(::BioSymbol)` for faster IO when dealing with symbols represented in text by ASCII characters.

The latter function is type piracy. For v3, I'd like to move the functionality of the latter method of `stringbyte` (perhaps with a less awful name) to `BioSymbols`.

jakobnissen:
Fix #198, companion PR to https://github.com/BioJulia/BioSymbols.jl/pull/48. See the PR to BioSymbols.

This change is not breaking. The changed function was undocumented and unexported.

:warning: This depends on BioSymbols 5.1.0 (more specifically PR 48) which is not yet released. :warning:

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,17 +9,17 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Twiddle = "7200193e-83a8-5a55-b20d-5d36d44a0795"
 
 [compat]
-BioSymbols = "5.0.0"
+BioSymbols = "5.1.0"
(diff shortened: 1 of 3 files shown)


jakobnissen:
Merging this - the failed doc build was due to a pkg server issue, I think.

PR #203, opened 2021-10-30
WIP: Test bonanza for v3

### LINKED ISSUE #139: Fixup tests before v3
Two things I'd like to improve on

* Some of my most recent changes here on version 2 have had insufficient testing. Ideally, I'd like to get coverage up near 100% again. This is much more easily achieved AFTER #138 , so this might be one of the later issues to address before v3
* There are way too many tests for BioSequences - nearly 9000 for LongSequence and 26000 for Mers. That's in package with 10k lines! On my (rather fast) laptop, they take 1.5 minutes. Not an eternity, but still a bit of wait. Worse, most of them don't really do anything except burn CI time and spam up the terminal if something goes wrong. I bet at least 75% of these tests could be eliminated if we cut down on test patterns, like
```
    for len in 1:64, _ in 1:10
        @test kmer_stuff(len, rand_seq(len))
    end
```
Do we really need 640 distinct tests here? What are the odds that function fails at k=54, and not before?

So, plan of action:
 - [x] Complete #138 , removing everything we don't want in this package
 - [x] Complete #121 , and any other new features to mers or longsequences
 - [x] Aggressively cut down on needless tests for mers and longsequences
 - [x] Add tests back to get near-100% coverage

Using [Aqua](https://github.com/JuliaTesting/Aqua.jl) on Julia exposes a few method ambiguities. In general, BioSequences have [an issue with method ambiguities](#135 ).

- [x] Fix method ambiguities (after all v3 feature issues, the possible interface rewrite, and #135 )

jakobnissen:
# Improve test coverage for v3

Resolves #139 . During the testing, it is possible some functions may be changed if they are discovered to be broken. This PR will contain an exhaustive list of all non-test changes, that is, changes in the `src` directory,

**Completion**
- [x] Alphabet
- [x] Translation
- [x] BioSequence
- [x] LongSequence
- [x] Bit-manipulation
- [ ] Search  

## Current changes / bugfixes
* `join!` is now exported as it should have been
* Fixed a typo in `join!(::BioSequence, ::Vector)` which caused it to fail
* Renamed `_copyto!` for `BioSequence` to `Base.copyto!`. I don't know why it was an internal method before, honestly.
* Remove `unsafe_setindex!`, which has been kept until now for backwards compatibility
* Fix typo `is_canonical` in `canonical(::BioSequence)`
* Remove outdated `Base.cmp` for `BitIndex`
* Remove unused `extract_encoded_element(::BitIndex, ::NTuple)` and `bitmask(::Type, ::Val)`
* Remove meaningless and non-tested function `LongSequence(it) = LongSequence{eltype(it)}(it)` (was also wrong)
* Changed un-tested constructor `LongSequence{A}(::Union{AbstractString,AbstractVector{UInt8}}, ::Integer, ::Integer)` to take a range instead of two integers for consistency with other constructors
* Remove one copy of duplicated method `bitindex(::LongSubSeq, ::Integer)`
* Remove one copy of duplicated method `encoded_setindex!(::SeqOrView, ::Unsigned, ::Integer)`
* Added missing method `symbols_per_data_element(::LongSubSeq)`
* Change various methods to take `AbstractUnitRange` over `UnitRange` with identical behaviour

### DIFF
--- a/docs/src/construction.md
+++ b/docs/src/construction.md
@@ -306,3 +306,16 @@ So the take home message of sequence literals is this:
 Be careful when you are using sequence literals inside of functions, and inside
 the bodies of things like for loops. And if you use them and are unsure, use the
  's' and 'd' flags to ensure the behaviour you get is the behaviour you intend.
+
+## Comparison to other sequence types
+Following Base standards, BioSequences do not compare equal to other containers even if they have the same elements.
+To e.g. compare a BioSequence with a vector of DNA, compare the elements themselves:
+```jldoctest
+julia> seq = dna"GAGCTGA"; vec = collect(seq);
+
+julia> seq == vec, isequal(seq, vec)
+(false, false)
+
+julia> length(seq) == length(vec) && all(i == j for (i, j) in zip(seq, vec))
+true 
+```
--- a/src/BioSequences.jl
+++ b/src/BioSequences.jl
@@ -126,6 +126,7 @@ export
     reverse_complement!,
     ungap,
     ungap!,
+    join!,
     
     ###
     ### LongSequence
--- a/src/biosequence/biosequence.jl
+++ b/src/biosequence/biosequence.jl
@@ -63,7 +63,7 @@ end
 
 # Fast path for iterables we know are stateless
 function join!(seq::BioSequence, it::Union{Vector, Tuple, Set})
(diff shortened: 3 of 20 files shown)


jakobnissen:
Since I consider this PR to be low-risk and unsurprising, I'm just merging right away, You're all welcome to review

PR #218, opened 2022-01-10
Add `findall` method

### LINKED ISSUE #217: Implement `findall`
We don't have this function, but we might as well.
We might want to not return a `Vector` of results but a lazy iterator though (what's up with Base needlessly materializing everything?!)

CiaranOMara:
I reckon this PR fulfils the `findall` method requirements. 

The `findall` method delegates to a `Search` iterator in the implementation. Let me know whether this approach suits and I'll add some docstrings.

Closes #217.

### DIFF
--- a/src/BioSequences.jl
+++ b/src/BioSequences.jl
@@ -11,7 +11,7 @@ export
     ###
     ### Symbols
     ###
-    
+
     # Types & aliases
     NucleicAcid,
     DNA,
@@ -81,7 +81,7 @@ export
     AA_X,
     AA_Term,
     AA_Gap,
-    
+
     # Predicates
     isGC,
     iscompatible,
@@ -90,35 +90,35 @@ export
     isgap,
     ispurine,
     ispyrimidine,
-    
+
     ###
     ### Alphabets
     ###
-    
+
     # Types & aliases
     Alphabet,
     NucleicAcidAlphabet,
     DNAAlphabet,
     RNAAlphabet,
     AminoAcidAlphabet,
-    
+
     ###
     ### BioSequences
     ###
 
     join!,
-    
+
     # Type & aliases
     BioSequence,
     NucSeq,
     AASeq,
-    
+
     # Predicates
     ispalindromic,
     hasambiguity,
     isrepetitive,
     iscanonical,
-    
+
     # Transformations
     canonical,
     canonical!,
@@ -129,11 +129,11 @@ export
     ungap,
     ungap!,
     join!,
-    
+
     ###
     ### LongSequence
     ###
-    
+
     # Type & aliases
     LongSequence,
     LongNuc,
@@ -141,7 +141,7 @@ export
     LongRNA,
     LongAA,
     LongSubSeq,
-    
+
     # Random
     SamplerUniform,
     SamplerWeighted,
@@ -149,18 +149,18 @@ export
     randdnaseq,
     randrnaseq,
     randaaseq,
-    
+
     ###
     ### Sequence literals
     ###
-    
+
     @dna_str,
     @rna_str,
     @aa_str,
 
     @biore_str,
     @prosite_str,
-    
+
     BioRegex,
     BioRegexMatch,
     matched,
@@ -177,7 +177,7 @@ export
     translate!,
     translate,
     ncbi_trans_table,
-    
+
     # Search
     ExactSearchQuery,
     ApproximateSearchQuery,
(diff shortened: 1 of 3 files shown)


src/BioSequences.jl
```diff
@@ -227,4 +227,46 @@ include("search/ApproxSearchQuery.jl")
...
+
+const HasRangeEltype = Union{<:ExactSearchQuery, <:ApproximateSearchQuery, <:Regex}
+
+Base.eltype(::Type{Search{Q,I}}) where {Q<:HasRangeEltype,I} = UnitRange{Int}
+Base.IteratorEltype(::Type{Search{Q,I}}) where {Q<:HasRangeEltype,I} = Base.IteratorEltype(UnitRange{Int})
+
+Base.eltype(::Type{<:Search}) = Int
+Base.IteratorEltype(::Type{<:Search}) = Base.IteratorEltype(Int)
+
+Base.IteratorSize(::Type{<:Search}) = Base.SizeUnknown()
```
jakobnissen, line 260:
```suggestion
Base.eltype(::Type{Search{<:Q}}) where {Q<:HasRangeEltype} = UnitRange{Int}
Base.eltype(::Type{<:Search}) = Int
Base.IteratorSize(::Type{<:Search}) = Base.SizeUnknown()
```

jakobnissen: `Base.IteratorSize` just needs to be `HasEltype()`, but this is already the default.

src/BioSequences.jl
```diff
@@ -227,4 +227,46 @@ include("search/ApproxSearchQuery.jl")
...
+
+function Base.findall(pat, seq::BioSequence, start::Integer=firstindex(seq); overlap::Bool = default_overlap, stop::Integer = lastindex(seq))
```
jakobnissen, line 262: I'm ambivalent about this signature. Why is start part of the signature but stop isn't? Might be good though, just not sure.

src/BioSequences.jl
```diff
@@ -227,4 +227,46 @@ include("search/ApproxSearchQuery.jl")
...
+
+struct Search{Q,I}
+    query::Q
+    itr::I
+    start::Integer
+    stop::Integer
+    overlap::Bool
+end
+
+const default_overlap = true
```
jakobnissen, line 240:
```suggestion
const DEFAULT_OVERLAP = true
```

src/BioSequences.jl
jakobnissen, line 262:
```suggestion
function Base.findall(pat, seq::BioSequence, start::Integer=firstindex(seq); overlap::Bool = DEFAULT_OVERLAP, stop::Integer = lastindex(seq))
```

src/BioSequences.jl
```diff
@@ -227,4 +227,46 @@ include("search/ApproxSearchQuery.jl")
 include("search/re.jl")
 include("search/pwm.jl")
 
+
+
+struct Search{Q,I}
+    query::Q
+    itr::I
+    start::Integer
+    stop::Integer
```
jakobnissen, line 236:
```suggestion
    start::Int
    stop::Int
```

jakobnissen (commented):
Great, thanks for contributing. Not sure about the signature though, it doesn't match `Base`. What signature would you prefer?

src/BioSequences.jl
jakobnissen: I think I would prefer having the same signature as `Base.findall(::AbstractString, ::AbstractString; overlap=true)` - though `Base` has overlap default to `false`, that seems strange to me.

If the user wants to limit the start and stop, they can create a view of the sequence.

src/BioSequences.jl
CiaranOMara: @jakobnissen, do we need the `<:Q`?
```julia
Base.eltype(::Type{<:Search{<:Q}}) where {Q<:HasRangeEltype} = UnitRange{Int}
vs
Base.eltype(::Type{<:Search{Q}}) where {Q<:HasRangeEltype} = UnitRange{Int}
```

jakobnissen: Ah no, you are right, we don't

CiaranOMara:
This PR is ready for another review pass.

I've added a convenience method that uses a view to limit the search range and map indices back to a parent. If it's not worthwhile, it can be removed.
https://github.com/BioJulia/BioSequences.jl/blob/fd0f57fa26abe0beffc05aa9bbadad0bac426e6c/src/BioSequences.jl#L259-L263

Also, is the PR code positioned where you'd like it to be within the project?

src/BioSequences.jl
CiaranOMara: An overlap default of `true` makes more sense to me in the context of BioSequences. But I'm okay changing to be consistent with `Base` if that's the consensus. However, setting the default to `false` may introduce a risk of missing important sequences.

jakobnissen: I agree - `true` is the most sensible default here 👍

jakobnissen (approved):

jakobnissen:
I think everything looks good now.

SabrinaJaye:
Looking good! Did you want to add some docstrings and or a paragraph in the search section of the manual? No immediate pressure as they can always be added before a release.

CiaranOMara:
I reckon I can add some docstrings to this over the next couple of days.

jakobnissen:
I'll just merge this, and then we can always add some docs later.

PR #220, opened 2022-01-29
Add `findall` method

CiaranOMara:
This PR:
- adds a docblock to the findall method.
- expands find examples where possible.
- adds doctests to unit tests.
- includes minor doctest corrections.
- enables `checkdocs` for build docs.

### DIFF
--- a/src/BioSequences.jl
+++ b/src/BioSequences.jl
@@ -11,7 +11,7 @@ export
     ###
     ### Symbols
     ###
-    
+
     # Types & aliases
     NucleicAcid,
     DNA,
@@ -81,7 +81,7 @@ export
     AA_X,
     AA_Term,
     AA_Gap,
-    
+
     # Predicates
     isGC,
     iscompatible,
@@ -90,35 +90,35 @@ export
     isgap,
     ispurine,
     ispyrimidine,
-    
+
     ###
     ### Alphabets
     ###
-    
+
     # Types & aliases
     Alphabet,
     NucleicAcidAlphabet,
     DNAAlphabet,
     RNAAlphabet,
     AminoAcidAlphabet,
-    
+
     ###
     ### BioSequences
     ###
 
     join!,
-    
+
     # Type & aliases
     BioSequence,
     NucSeq,
     AASeq,
-    
+
     # Predicates
     ispalindromic,
     hasambiguity,
     isrepetitive,
     iscanonical,
-    
+
     # Transformations
     canonical,
     canonical!,
@@ -129,11 +129,11 @@ export
     ungap,
     ungap!,
     join!,
-    
+
     ###
     ### LongSequence
     ###
-    
+
     # Type & aliases
     LongSequence,
     LongNuc,
@@ -141,7 +141,7 @@ export
     LongRNA,
     LongAA,
     LongSubSeq,
-    
+
     # Random
     SamplerUniform,
     SamplerWeighted,
@@ -149,18 +149,18 @@ export
     randdnaseq,
     randrnaseq,
     randaaseq,
-    
+
     ###
     ### Sequence literals
     ###
-    
+
     @dna_str,
     @rna_str,
     @aa_str,
 
     @biore_str,
     @prosite_str,
-    
+
     BioRegex,
     BioRegexMatch,
     matched,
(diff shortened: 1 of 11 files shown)


CiaranOMara:
@SabrinaJaye, @jakobnissen, if these docs are good enough, could you make this the new HEAD?

CiaranOMara:
@SabrinaJaye, could this be made the new HEAD?

src/BioSequences.jl
```diff
@@ -227,4 +227,83 @@ include("search/ApproxSearchQuery.jl")
 include("search/re.jl")
 include("search/pwm.jl")
 
+
```
SabrinaJaye, line 230: Can this be put in the search folder instead of the top level module file?

SabrinaJaye (changes requested):

src/BioSequences.jl
CiaranOMara: I've put it into the search folder. Let me know if you had different else in mind.

src/search/search.jl
```diff
@@ -0,0 +1,76 @@
...
+
+Find all occurrences of `pattern` in `sequence`.
+
+The return value is a vector of ranges of indices where the matching sequences were found.
+If there are no matching sequences, the return is an empty array.
+
+The search is restricted to the specified range when `rng` is set.
+
+With the keyword argument `overlap` set as `true`, the start index for the next search is set to the start of the current match plus one; if set to `false`, the start index for the next search is set to the end of the current match plus one.
+The default value for the keyword argument `overlap` is `true`.
```
jakobnissen, line 35:
```suggestion
With the keyword argument `overlap` set as `true`, the start index for the next search is set to
the start of the current match plus one; if set to `false`, the start index for the next search is
set to the end of the current match plus one.
The default value for the keyword argument `overlap` is `true`.
```

src/search/search.jl
```diff
@@ -0,0 +1,76 @@
...
+Base.eltype(::Type{<:Search}) = Int
+Base.IteratorSize(::Type{<:Search}) = Base.SizeUnknown()
+
+"""
+    findall(pattern, sequence::BioSequence[,rng::UnitRange{Int}]; overlap::Bool=true)
+
+Find all occurrences of `pattern` in `sequence`.
+
+The return value is a vector of ranges of indices where the matching sequences were found.
+If there are no matching sequences, the return is an empty array.
```
jakobnissen, line 30:
```suggestion
    findall(pattern, sequence::BioSequence[,rng::UnitRange{Int}]; overlap::Bool=true)::Vector

Find all occurrences of `pattern` in `sequence`.

The return value is a vector of ranges of indices where the matching sequences were found.
If there are no matching sequences, return an empty array.

`pattern` can be a `Biosymbol` or a predicate.
```

test/longsequences/find.jl
```diff
@@ -28,4 +28,26 @@
...
+    @test findall(ExactSearchQuery(dna"GATC", iscompatible), seq; overlap = true) == [1:4, 2:5, 6:9, 10:13]
+
+    # Check mapping of indices.
+    @test findall(DNA_A, seq, 7:11) == [7, 11]
+    @test findall(ExactSearchQuery(dna"A"), seq, 7:11) == [7:7, 11:11]
+
+    # Check empty return type.
+    @test findall(DNA_A, dna"GGGG") |> typeof == Vector{Int}
+    @test findall(ExactSearchQuery(dna"A"), dna"GGGG") |> typeof == Vector{UnitRange{Int}}
+
```
jakobnissen, line 52: Need tests for searching for predicate. See your comments in #235

src/search/search.jl
```diff
@@ -0,0 +1,76 @@
...
+2-element Vector{UnitRange{Int64}}:
+ 1:4
+ 5:8
+
+julia> findall(ExactSearchQuery(dna"ACAC"), sequence, 2:7; overlap=false)
+1-element Vector{UnitRange{Int64}}:
+ 3:6
+```
+"""
+function Base.findall(pattern, sequence::BioSequence; overlap::Bool = DEFAULT_OVERLAP)
```
jakobnissen, line 68: Ambiguous with `Base.findall(::Function, ::Any)` See #235

jakobnissen (changes requested):
Thanks for this PR. Adding doctests is a good contribution, I see you found quite a lot of bugs.
I do have some slight wording issues in the docstring, and also the issues from #235 needs to be addressed.

CiaranOMara:
Rebasing this PR was taking too long.

PR #221, opened 2022-01-30
Loosen search iterator

CiaranOMara:
This PR loosens the `Search` iterator to accommodate iterating in forward and reverse directions.

It needs a little more work to support the `ApproxSearchQuery`. It wasn't immediately obvious how to have the iterator generically include or pass through additional arguments to an underlying function. The other approach is to have `ApproxSearchQuery` own the additional argument for the number of allowed errors  - comments would be helpful. 

Let me know whether this seems like a good direction for a search iterator.

### DIFF
--- a/src/BioSequences.jl
+++ b/src/BioSequences.jl
@@ -229,7 +229,8 @@ include("search/pwm.jl")
 
 
 
-struct Search{Q,I}
+struct Search{F,Q,I}
+    findfunc::F
     query::Q
     itr::I
     overlap::Bool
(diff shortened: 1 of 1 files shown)


SabrinaJaye:
That seems very desirable to me, I remember thinking looking at the search code - "hmm how would you make Approx go backwards?".

CiaranOMara:
The `Search` iterator would become generic or complete in its current state if all the queries owned their additional arguments. However, changing `ApproxSearchQuery` is a breaking change. So a design discussion is needed.

I think we would be in a better position if we constrain our `findnext` and `findprev` method signatures to the following.
```julia

function findnext(query, seq::BioSequence, start::Integer)
    ...
end

function findprev(query, seq::BioSequence, start::Integer)
    ...
end
```

All the peripheral `findfirst`, findlst`, `findall` methods can become generic if we adopt this constraint.

In terms of the `ApproxSearchQuery`, the additional `k` argument would need adopting, and then another wrapper/layer may be required to optimise for the reuse of `ApproxSearchQuery`'s work.

PR #228, opened 2022-05-31
Fix small errors

jakobnissen:
These errors were caught by JET's automatic analysis of packages (it's awesome you guys).
`GeneticCode` is immutable and need not support `copy`, but it's kept as a noop for compatibility.
`Regex` really needs one of these three types, the code assumes them, so we might as well enforce it at compile time.

### DIFF
--- a/src/geneticcode.jl
+++ b/src/geneticcode.jl
@@ -33,7 +33,7 @@ function Base.getindex(code::GeneticCode, codon::UInt64)
     return @inbounds code.tbl[codon + one(UInt64)]
 end
 
-Base.copy(code::GeneticCode) = GeneticCode(copy(code.name), copy(code.tbl))
+Base.copy(code::GeneticCode) = code
 Base.length(code::GeneticCode) = 64
 
 Base.show(io::IO, code::GeneticCode) = print(io, code.name)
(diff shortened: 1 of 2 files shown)


kescobo (approved):
Cool! LGTM. Does it make sense to add a test that `copy(code::GeneticCode) === code`?

SabrinaJaye (approved):

CiaranOMara:
I like how JET found a way to make `show` determinate. However, it left a redundant comparison. Though, maybe the comparison gets optimised out and is not seen?
https://github.com/BioJulia/BioSequences.jl/blob/26e05f723040bd6726ec1e9be37332fec1c4abab/src/search/re.jl#L619

Maybe the section could be refactored into the following?
```julia
function descriptor(::Type{BioSequences.RNA})
    return "rna"
end

function descriptor(::Type{BioSequences.DNA})
    return "dna"
end

function descriptor(::Type{BioSequences.AminoAcid})
    return "aa"
end

function Base.show(io::IO, re::Regex{T}) where {T}
    opt = descriptor(T)
    print(io, "biore\"", re.pat, "\"", opt)
end
```

The descriptors are used as postfixes or prefixes in other places, if I recall correctly.

PR #229, opened 2022-06-07
Sequence hashing: Bugfix and slight simplification

jakobnissen:
Slightly simplify hashing code and fix a minor bug:
* Replace manual bitrotate code with Base.bitrotate (from Julia 1.5)
* Fix an error from a typo in in murmur2 function
* Slightly simplify the code
* Improve test coverage

This results in a ~12% speed increase on my computer for large sequences, but 9% decrease for short sequences. Might just be measurement noise.

### DIFF
--- a/src/longsequences/hash.jl
+++ b/src/longsequences/hash.jl
@@ -18,10 +18,6 @@
 const c1 = 0x87c37b91114253d5
 const c2 = 0x4cf5ad432745937f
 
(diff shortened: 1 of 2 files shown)


jakobnissen:
Merging this, as it's a low-risk change with no user-facing changes, other than the precise hash value has changed, which is fair game considering it's a bugfix.

Issue #219, opened 2022-01-20
add an `ispermitted` for Alphabets?


TransGirlCodes:
Can we add an `ispermitted` method for Alphabets?

So the background context for this is I'm doing some of the internals for Kmers.jl - iterators being the last thing to move over and make nice. An recall when kmers only supported 2 bit alphabets, iterating over kmers for the generic method where a Kmer and the sequence kmers are generated from having differing alphabets, then gaps or anything that is not in the 2 bit alphabet, would be skipped over. In constructors it would throw an error - makes sense - the user is asking to make a kmer from something very specific, whereas we want the iterator to just drop the weird symbols and keep going. With the encode and decode methods as they are, they throw, but for my purpose of writing Kmer.jl internals, it would be really useful to have `encode` and `decode`
basically work like `encode(::A, x) = ispermitted(A(), x) ? unsafe_encode(A(), x) : throw_n_stuff`, so I can use the ispermitted and unsafe_encode methods manually and have the iterator do the right thing rather than cause a throw.

CiaranOMara:
I accept that there could be weird symbols. However, I'm not sure I understand the encoding and decoding of a weird symbol, the necessary constraints that would allow valid symbols to continue functioning, and what `KMER` is constructed from in your scenario. I think there are two construction scenarios, one from a structure with bit packing and one without (Vector). 

I think there are more constraints to consider when there is bit packing. For example, would there be a constraint on `unsafe_encode` such that it must encode with the same number of bits? Though I assume, there are not necessarily enough bits to represent all the uniquely weird symbols that may be encountered. So I think this assumption pushes us towards having something in BioSequences that represents the presence of an invalid symbol, which would drop information. What do you think?

In terms of the skipping over that you spoke of, does that mean that the next valid symbol gets included in the KMER?

jakobnissen:
So if I understand it correctly, you want a `tryencode(::Alphabet, x)::Union{eltype(A), nothing}`? That sounds reasonable.
Should the title of this issue be changed?

jakobnissen:
If what you are literally asking is a way to check if some symbol is allowed in an `Alphabet`, this can already be done with `x in symbols(A)`. This is not optimised, though, so this method could be possible if `tryencode` is not sufficient (although it really should be!)

jakobnissen:
Closed by #324

PR #238, opened 2022-06-16
Add tryencode and trydecode

jakobnissen:
# Add tryencode and trydecode

The encode and decode methods are not allowed to produce invalid data. Instead,
they throw an error when encountering invalid input data.
This can lead to some frustration when checking if a symbol is permitted in an
alphabet.
One way to solve it is by checking `symbol in symbols(A)`, but this is not
particularly effective.

This PR adds a tryencode and trydecode method to existing alphabets. These
methods return nothing when given invalid data.
Methods encode and decode now internally call their try-variants.

May solve #219

## Types of changes

* `tryencode` and `trydecode` introduced, returning `Union{Nothing, E}`
* `encode` and `decode` now uses a fallback implementation which calls these try- methods internally
* There is no performance difference
* Alphabet docstring has been changed in a backwards compatible way to mention these new methods.

- [X] :blue_book: I have updated and added relevant docstrings, in a manner consistent with the [documentation styleguide](https://docs.julialang.org/en/v1/manual/documentation/).
- [X] :blue_book: I have added or updated relevant user and developer manuals/documentation in `docs/src/`.
- [X] :ok: There are unit tests that cover the code changes I have made.
- [X] :ok: The unit tests cover my code changes AND they pass.
- [X] :pencil: I have added an entry to the `[UNRELEASED]` section of the manually curated `CHANGELOG.md` file for this repository.
- [X] :ok: All changes should be compatible with the latest stable version of Julia.
- [X] :thought_balloon: I have commented liberally for any complex pieces of internal code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 * It is now possible to `join` BioSymbols into a BioSequence.
+* Added non-exported functions `trydecode` and `tryencode`.
 
 ## [3.0.1]
 ### Removed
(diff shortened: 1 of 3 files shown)


jakobnissen:
CodeCov is acting up. There is no code coverage difference as far as I can see.

jakobnissen:
Superseded by #324

PR #242, opened 2022-06-16
Schedule method for removal

CiaranOMara:
I'm unsure whether this suggestion is a good way to go about planned deprecation. The idea is that if removal is forgotten or missed, the method won't occur in the v2 API. What do you think?

cc: @jakobnissen

### DIFF
--- a/src/longsequences/constructors.jl
+++ b/src/longsequences/constructors.jl
@@ -46,7 +46,7 @@ Base.empty(::Type{T}) where {T <: LongSequence} = T(UInt[], UInt(0))
 
 # Constructors from other sequences
 
-@static if VERSION < v"2"
+@static if VERSION < v"4"
     # TODO: Remove this method, since the user can just slice
     LongSequence(s::LongSequence, xs::AbstractUnitRange{<:Integer}) = s[xs]
 end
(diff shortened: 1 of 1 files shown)


jakobnissen:
Good idea - but we're already on V3, so it should be `@static if VERSION < v"4"` :)

src/longsequences/constructors.jl
```diff
@@ -45,9 +45,12 @@ Base.empty(::Type{T}) where {T <: LongSequence} = T(UInt[], UInt(0))
 (::Type{T})() where {T <: LongSequence} = empty(T)
 
 # Constructors from other sequences
-# TODO: Remove this method, since the user can just slice
-LongSequence(s::LongSequence, xs::AbstractUnitRange{<:Integer}) = s[xs]
 
+@static if VERSION < v"2"
```
jakobnissen, line 49:
```suggestion
@static if VERSION < v"4"
```

jakobnissen (changes requested):

jakobnissen:
Wait no, this doesn't make any sense, sorry. `VERSION` refers to Julia's version. We can remove the method in the next breaking release of BioSequences, regardless of Julia version.
We can't access BioSequence's own version from within the package, unless we parse our own Project.toml at compile time, which is annoying and error prone. Worth it for some important change, but not for this, I think.

CiaranOMara:
Oh dear, my morning coffee mustn’t have taken effect.

Issue #243, opened 2022-07-05
Design of hash(::BioSequence). What to do?


jakobnissen:
So the design of BioSequences makes it difficult to implement efficient and correct hashing.
We want efficient hashing, because hashing underlies operations like putting stuff in a `Set` or `Dict`, which users expect to be fast.

The issue is
* Julia requires that `isequal(a, b) === isequal(hash(a, x), hash(b, x))`
* `isequal` ought to allow objects of different types to be equal if it represents the same value. E.g. `isequal(0, 0.0)`. Breaking this leads to lots of confusion.

So, to follow these two rules
* Different subtypes of `BioSequence` should be `isequal` if they have the same content, i.e. `isequal(dna"TAG", DNAKmer("TAG"))`
* Which implies they should hash equally,

Now, how do we get two `BioSequence`s with arbitrary encoding to hash equivalently? As I see it, it means we can't hash the encoded data, because the encoded data may vary between subtypes. However, this is presumably the only way to avoid decoding, which is the only way to make hashing fast!

Incidentally, the current implementation of `hash` for `LongSequence` is broken:
```
julia> (a, b) = (LongDNA{4}("A"), LongDNA{2}("A"));

julia> a in [b] # because they are equal
true

julia> a in Set([b]) # hashes wrong
false
```

Here are some possible solutions:
1. Just implement hashing by hashing every element. This will make hashing significantly slower, but it will be simple and correct.
2. Make one encoding privileged, e.g. `LongSequence`'s. When hashing any other type, we re-code it to that encoding before hashing. This will keep LongSequence hashing fast, but will make everything else both slower and much more complex. If we do this, we also need to recode `LongSequence{NucleicAcidAlphabet{2}}`.

There may be other solutions.

kescobo:
>Incidentally, the current implementation of hash for LongSequence is broken

:grimacing: This is not good. We might want a separate issue to get a bug-fix on that specifically.

>Here are some possible solutions:

Regarding this, I have a question - is changing the content of a hash breaking, or only the behavior. That is, as long as we fulfill the hashing contract(s), if we change the underlying behavior to improve speed, is that breaking?

If the answer is no, I propose that we do (1) as a stop-gap to fix the current bug, since it's simple, and then we can work on further optimization separately. 

> `isequal` ought to allow objects of different types to be equal if it represents the same value. E.g. isequal(0, 0.0). Breaking this leads to lots of confusion.

[...]

PR #244, opened 2022-07-05
Use naive hash fallback for BioSequence

jakobnissen:
Current hashing behaviour hashes equal biosequences with different encodings to
different values, which violate the interface of hash.
This is a fix until (and if) we figure out a better method of hashing.

See issue #243

Going to immediately release this as a new version if tests pass, cc @kescobo @SabrinaJaye


jakobnissen:
Codecov is an artifact of suddenly having less (tested) source code, meaning the untested fraction increases - this is completely covered by tests.

PR #250, opened 2022-07-15
Add rand!([::AbstractRNG], ::LongSequence, [::Sampler]) methods

jakobnissen:
I had apparently forgotten about those before.
Also refactor the old code a bit to make it easier to reason about.


jakobnissen:
This should be good to go. I'm merging as soon as tests pass - I can't imagine anyone would have anything against this change.

Issue #263, opened 2023-01-23
Translate subsequence 


jakobnissen:
Currently does not work, but should. In fact, translation should be generic over BioSequence

~~Point to consider: Should it return a LongSequence? If so, people can't override translate to their own custom types without piracy.~~
Nevermind, extending functions with own types is not piracy. See the PR for more discussion.

PR #264, opened 2023-01-29
Allow translation of sequence views

jakobnissen:
See #263 

This PR extends existing `translate` code to also work on `LongSubSeq`, which also returns `LongAA`.
Other subtypes of `BioSequence` must implement their own version of `translate`.

This also changes how `GeneticCode` is indexed, which is technically breaking. Previously, it used an unsafe `@inbounds` index using an `UInt64`, zero-indexed. Now, it's generic over `Integer`, guarded behinds a `@boundscheck`, and one-based, as is idiomatic. I think it's unlikely someone relied on the old, unsafe behaviour.

## TODO
* ~~It would be nicer if the implementation was generic across sequence types, but I'm not sure how to do this. Perhaps we could make it so sequences of type `Seq{Alph}` creates a `Seq{AminoAcidAlphabet}`, then filled it in. This would only work for mutable sequences.~~ Nevermind, I can't do that, because subtypes of BioSequence may not be parameterized.
* This needs tests. It would be nice to refactor the translation tests now that I'm at it.

## Types of changes
* [X] :boom: Breaking change (fix or feature that would cause existing functionality to change).

- [x] :blue_book: I have updated and added relevant docstrings, in a manner consistent with the [documentation styleguide](https://docs.julialang.org/en/v1/manual/documentation/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [UNRELEASED]
+### Added
+* `LongSubSeq{<:NucleicAcidAlphabet}` can now be translated.
 
 ## [3.1.0]
 ### Added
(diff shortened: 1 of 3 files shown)


camilogarciabotero:
Hi all,

Is this PR working well? I was wondering if after merging it, the `LongNucOrView{N}` will be exported as well, I have found myself trying to apply methods to either a `LongSeq` or `LongSubSeq`.

Best.

PR #271, opened 2023-03-01
Fix behavior of iterate on indices <= 0

cjprybol:
## Types of changes

* [ ] :sparkles: New feature (A non-breaking change which adds functionality).
* [x] :bug: Bug fix (A non-breaking change, which fixes an issue).

This is how iterate behaves on strings
```julia
julia> x = "ATCG"
"ATCG"

julia> iterate(x)
('A', 2)

julia> iterate(x, -1)

julia> iterate(x, 0)

julia> iterate(x, 1)
('A', 2)

julia> iterate(x, 2)
('T', 3)

julia> iterate(x, 4)
('G', 5)

julia> iterate(x, 5)
```

iterate currently behaves as expected when indexing >= length(seq), but will crash when using any index less than or equal to the initial position (1)
```julia
julia> x = BioSequences.randdnaseq(3)
3nt DNA Sequence:
TGG

julia> iterate(x, 4)

julia> iterate(x, 0)

signal (11): Segmentation fault
in expression starting at REPL[5]:1
>> at ./int.jl:456 [inlined]
extract_encoded_element at /home/jovyan/workspace/BioSequences.jl/src/bit-manipulation/bitindex.jl:74 [inlined]
extract_encoded_element at /home/jovyan/workspace/BioSequences.jl/src/longsequences/indexing.jl:18 [inlined]
getindex at /home/jovyan/workspace/BioSequences.jl/src/biosequence/indexing.jl:39 [inlined]
iterate at /home/jovyan/workspace/BioSequences.jl/src/biosequence/indexing.jl:11
unknown function (ip: 0x7f5e879a9631)
```

- [ ] :art: The changes implemented is consistent with the [julia style guide](https://docs.julialang.org/en/v1/manual/style-guide/).
- [ ] :blue_book: I have updated and added relevant docstrings, in a manner consistent with the [documentation styleguide](https://docs.julialang.org/en/v1/manual/documentation/).

## Notes

Found this while trying to fix https://github.com/BioJulia/Kmers.jl/issues/33 - I think I've got that issue fixed, but this (unrelated issue) was the next failing test that I hit in the Kmers.jl tests

### DIFF
--- a/src/biosequence/indexing.jl
+++ b/src/biosequence/indexing.jl
@@ -8,7 +8,7 @@
 ### License is MIT: https://github.com/BioJulia/BioSequences.jl/blob/master/LICENSE.md
 
 @inline function Base.iterate(seq::BioSequence, i::Int = firstindex(seq))
-    i > lastindex(seq) ? nothing : @inbounds seq[i], i + 1
+    (i <= 0) || (i > lastindex(seq)) ? nothing : @inbounds seq[i], i + 1
 end
 
 ## Bounds checking
--- a/src/biosequence/indexing.jl
+++ b/src/biosequence/indexing.jl
@@ -8,7 +8,7 @@
 ### License is MIT: https://github.com/BioJulia/BioSequences.jl/blob/master/LICENSE.md
 
 @inline function Base.iterate(seq::BioSequence, i::Int = firstindex(seq))
-    (i % UInt) - 1 < lastindex(seq) ? (@inbounds seq[i], i + 1) : nothing
+    (i % UInt) - 1 < (lastindex(seq) % UInt) ? (@inbounds seq[i], i + 1) : nothing
 end
 
 ## Bounds checking
--- a/src/biosequence/indexing.jl
+++ b/src/biosequence/indexing.jl
@@ -8,7 +8,7 @@
 ### License is MIT: https://github.com/BioJulia/BioSequences.jl/blob/master/LICENSE.md
 
 @inline function Base.iterate(seq::BioSequence, i::Int = firstindex(seq))
-    (i <= 0) || (i > lastindex(seq)) ? nothing : @inbounds seq[i], i + 1
+    (i % UInt) - 1 < lastindex(seq) ? (@inbounds seq[i], i + 1) : nothing
 end
 
 ## Bounds checking
--- a/test/longsequences/iteration.jl
+++ b/test/longsequences/iteration.jl
@@ -10,4 +10,13 @@
     aa_seq = aa"ARNPS"
     aa_vec = [AA_A, AA_R, AA_N, AA_P, AA_S]
     @test all([aa == aa_vec[i] for (i, aa) in enumerate(aa_seq)])
+    
+    @test iterate(dna_seq) == iterate(dna_seq, 1) == (DNA_A, 2)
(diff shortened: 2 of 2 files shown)


jakobnissen:
I'm not crazy about this change TBH. The reason is that the second argument to iterate (the state) is generally regarded as an internal object. Using iterate with your own state is, in general in Julia, undefined behaviour. For example:
```julia
julia> s = Set([1,2])
Set{Int64} with 2 elements:
  2
  1

julia> iterate(s, -10000000)
(1392277128300, -9999998)
```
So there is really no issue here to fix. And adding one more branch in a function as low-level as iterate does have some performance implications.

You _could_ change the iteration to have a similar implementation as that of `Array`, where it reinterprets to `UInt`, then exploits the fact that negative unsigned overflow in order to skip the extra check.

cjprybol:
Thanks for the feedback!

I was thinking of BioSequences being most similar to strings and/or arrays, and so those were the 2 types that I tried to mimic the behavior of
```julia
# strings are in the initial post above, arrays here
julia> x = [1]
1-element Vector{Int64}:
 1

julia> iterate(x, 0)

julia> iterate(x, 1)
(1, 2)

julia> iterate(x, 2)

```

I didn't try any other collections, but I see your point though that there isn't a universal standard for how to handle the scenario of iterating at state <= 0, and returning undefined behavior is just as valid as handling the bounds checking.

I like the idea of mimicking the reference implementation in Base as you mentioned:
```julia
julia> @enter iterate(x, 0)
[ Info: tracking Base
In iterate(A, i) at array.jl:781
>781  iterate(A::Array, i=1) = (@_inline_meta; (i % UInt) - 1 < length(A) ? (@inbounds A[i], i + 1) : nothing)
``` 

But just to confirm, when you say
> So there is really no issue here to fix.

Do you agree that we should fix the segfault crash?

If so, it sounds like you want to make sure the updates preserve performance rather than comply 100% with the behavior of arrays and strings. In which case trying the implementations in base, which have likely been optimized and benchmarked, is the best place to start

jakobnissen:
What I mean is that supplying your own state to iterate is undefined behaviour so it doesn't matter if it segfaults. However, since the implementation of iterate for Array manages to be safe with zero cost, we might as well use that, since there is no downside

src/biosequence/indexing.jl
```diff
@@ -8,7 +8,7 @@
 ### License is MIT: https://github.com/BioJulia/BioSequences.jl/blob/master/LICENSE.md
 
 @inline function Base.iterate(seq::BioSequence, i::Int = firstindex(seq))
-    i > lastindex(seq) ? nothing : @inbounds seq[i], i + 1
+    (i % UInt) - 1 < lastindex(seq) ? (@inbounds seq[i], i + 1) : nothing
```
jakobnissen, line 11:
```suggestion
    (i % UInt) - 1 < (lastindex(seq) % UInt) ? (@inbounds seq[i], i + 1) : nothing
```

jakobnissen: The UInt/Int comparison adds an extra branch. Don't know how arrays avoids this, but we can elide it like this

jakobnissen:
Thanks for your contribution!

PR #287, opened 2023-09-26
Remove downstream BioMakie CI

jakobnissen:
This has been consistently broken ever since it was introduced. Also, the tests take like 20 minutes before they fail, single-handedly multiplying BioSequences CI time by like 4.


kescobo:
Cc @kool7d

PR #293, opened 2023-10-16
Bump Julia compat to 1.6

jakobnissen:
We actually didn't support Julia 1.5, and didn't test on it, so it was a mistake to have claimed to have supported earlier versions.


jakobnissen:
I was wrong - it DOES run on Julia 1.5. See #294

PR #297, opened 2023-10-22
add  predicate method and tests

camilogarciabotero:
# Adds the new `hasprematurestop` method

Hi,

This is a simple new method that looks into a `BioSequence{<:NucleicAcidAlphabet}` and check if there is a premature stop codon. It only happens to run when there is an actual stop codon in the sequence, otherwise throws a `MethodError` with the corresponding message that no stop codon at the end. It has three test cases, I think they could cover all the instances.

I'm open to comments. This is my first PR in this  GH organization and hope the actual method is useful. I have used in one of my packages, but I think it could actually be nice if others get this functionality out from the `BioSequences` package.

## Types of changes

* [x] :sparkles: New feature (A non-breaking change which adds functionality).

- If you have implemented new features or behaviour
  - **Provide a description of the addition** in as many details as possible.

  - **Provide justification of the addition**.

  - **Provide a runnable example of use of your addition**. This lets reviewers
    and others try out the feature before it is merged or makes it's way to release.

- If you have changed current behaviour...
  - **Describe the behaviour prior to you changes**

  - **Describe the behaviour after your changes** and justify why you have made the changes,
    Please describe any breakages you anticipate as a result of these changes.

  - **Does your change alter APIs or existing exposed methods/types?**
    If so, this may cause dependency issues and breakages, so the maintainer
    will need to consider this when versioning the next release.

  - If you are implementing changes that are intended to increase performance, you
    should provide the results of a simple performance benchmark exercise
    demonstrating the improvement. Especially if the changes make code less legible.

- [x] :art: The changes implemented is consistent with the [julia style guide](https://docs.julialang.org/en/v1/manual/style-guide/).
- [x] :blue_book: I have updated and added relevant docstrings, in a manner consistent with the [documentation styleguide](https://docs.julialang.org/en/v1/manual/documentation/).

### DIFF
--- a/src/BioSequences.jl
+++ b/src/BioSequences.jl
@@ -118,6 +118,7 @@ export
     hasambiguity,
     isrepetitive,
     iscanonical,
+    hasprematurestop,
 
     # Transformations
     canonical,
--- a/src/biosequence/predicates.jl
+++ b/src/biosequence/predicates.jl
@@ -133,7 +133,7 @@ Determine whether the `sequence` of type `LongSequence{DNAAlphabet{4}}` contains
 
 Returns a boolean indicating whether the `sequence` has more than one stop codon.
 """
-function hasprematurestop(seq::LongSequence{DNAAlphabet{4}})::Bool
+function hasprematurestop(seq::BioSequence{<:NucleicAcidAlphabet})::Bool
     
     stopcodons = [dna"TAA", dna"TAG", dna"TGA"]  # Create a set of stop codons
     
--- a/src/biosequence/predicates.jl
+++ b/src/biosequence/predicates.jl
@@ -125,3 +125,28 @@ function iscanonical(seq::NucleotideSeq)
     end
     return true
 end
+
+"""
+    hasprematurestop(sequence::LongSequence{DNAAlphabet{4}})::Bool
+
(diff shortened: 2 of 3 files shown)


jakobnissen:
Thank you for your contribution.

I think the functionality in this PR is too application-specific and somewhat too idiosyncratic to belong in BioSequences. What I mean is that this function allows you to do a very specific query, that is probably not of broad usefulness. The implementation had be asking questions like:
* Why must there be a stop codon at the end? Wouldn't it be useful to have very similar functionality for checking if an ORF has any stop codons at all?
* What about different genetic codes, e.g. the human mitochondrial one?
* Why can't this function search for stop codons from the middle of a sequence, or from any arbitrary positions? That would be very useful when scanning for ORFs.

Me having these questions doesn't mean it's bad functionality. But it does probably mean that there is some more generic functionality that could be exposed to users, that would allow the users to build the functionality in this PR, plus address all my questions.

Maybe this functionalitiy should be something like
* Search for an amino acid in a nucleotide sequence, given a specific genetic code (with `find*` functions)
* Allow searches to only search in-frame, somehow.
* Implement `findnext`, `findlast` and `findall` for `BioRegex`.

This could also be useful to search for other motifs in sequences such as TATA boxes. I'm not sure what this API could look like, but it might be worth thinking about.

I'm closing this PR now, but I'll make an issue to discuss this.

PR #305, opened 2024-05-28
Precompute ExactSearchQuery faster

jakobnissen:
Compute bshift and fshift in one pass instead of two. This makes it a little faster to create exact search queries.


jakobnissen:
Tests fail because Julia 1.5 doesn't support MacOS on ARM - not an issue with this PR.

PR #311, opened 2024-10-17
Drop pre-LTS Julia support

jakobnissen:
This allows us to simplify CI, since we can run the oldest supported Julia version on aarch64 macos, and also allows us to mark const field const.


kescobo (approved):
This makes sense to me :+1:

PR #316, opened 2024-10-19
Rewrite counting of BioSequence

jakobnissen:
This is a large commit with multiple changes:

1. Remove bitparallel compiler
The previous bitparallel compiler was not great. It made use of metaprogramming,
which
    - Is hard to reason about
    - Does not play nice with Revise
    - Cannot be debugged using a debugger or similar tooling
    - Confuses linters and IDEs
    - May explode codegen, impacting compile times

This might have been necessary in earlier versions of Julia, but the modern
Julia compiler is so good at constant evaluation and inlining that this is
almost never a good idea.

2. Add internal chunk iterator types
This is an internal abstraction, an iterator that yields the encoded data chunks
of LongSequence or LongSubSeq. For subsequences, the yielded chunks are aligned,
such that the iterator returns the exact same elements as a chunk iterator
over the correponding LongSequence. These iterators might be generally useful in
future BioSequences work, and is something I've thought about for some time.

3. Deprecate the functions `n_ambigous`, `n_gaps` and `n_certain`
These should instead be accessed via their corresponding `count` methods, as
described in the documentation.
The rationale is that these functions provide little value as standalone
functions, as they should work exactly like `count` with specific predicates.
As we might want to add more in the future, this simply balloons the number
of exported symbols.

4. Deprecate some meaningless two-sequence counting methods
The method `n_ambiguous(::BioSequence ::BioSequence)`, and the equivalent for
n_gaps, n_certain and the `count` methods corresponding to these are strange
and obscure. Why are they there?
They significantly complicate the implementation of counting, while providing
no meaningful biological operations. At least I have no idea what the biological
significance of these methods are

5. Deprecate calling `matches` or `mismatches` with differing seq lengths
It is not clear what the correct thing to do here is, and it may be confusing
for users, who may expect a pairwise alignment or similar.
Instead of guessing what the user wants, force them to take a view of the longer
sequence if these differ in length.

6. Add new optimised methods for counting occurrences of symbol in sequence
E.g. `count(==(DNA_A), dna"TAG")` and a similar method for `isequal`.


jakobnissen:
@kescobo Added you as reviewer. This is a huge PR, so please don't feel the need to review all the bithacking. I'm more interested in how reasonable the changes to the API are - if you could focus on the changes made in the documentation (and that API changes those imply) that'd be grand.

docs/src/counting.md
```diff
@@ -6,66 +6,73 @@ end
 ```
 
 # Counting
+`BioSequences` contain functionality to efficiently count biosymbols in a biosequence
```
kescobo, line 9: If this is referring to the package, i think it should use the singular verb (eg `Biosequences.jl` contains") - if it refers to multiple instances of a type, it should maybe be `BioSequence`s (the "s" outside the ticks)

docs/src/counting.md
```diff
@@ -6,66 +6,73 @@ end
...
+    ns
+end 
+```
 
-BioSequences extends the `Base.count` method to provide some useful utilities for
-counting the number of sites in biological sequences.
-
-Most generically you can count the number of sites that satisfy some condition
-i.e. cause some function to return `true`:
+This function can be more efficient implemented by exploiting the internal
```
kescobo, line 23:
```suggestion
This function can be more efficiently implemented by exploiting the internal
```

docs/src/counting.md
```diff
@@ -6,66 +6,73 @@ end
...
+    It is important to understand that this speed is achieved with custom methods of
+    `Base.count`, and not by a generic mechanism that improves the speed of counting
+    symbols in `BioSequence `in general.
+    Hence, while `count(==(DNA_N), seq)` may be optimised,
+    `count(i -> i == DNA_N, seq)` is not, as this is a different method.
 
-```
+## Currently optimised methods
+By default, only `LongSequence` and `LongSubSeq` have optimised methods. Downstream
+implementers of new `BioSequence`s may not be optimised.
```
kescobo, line 37: This sort of sounds like you're saying they *can't* be optimized, but I think you're saying it's possible that they will, but also possible that they won't. 

"May not" in English is ambiguous - it sometimes means "might not" and sometimes means "not allowed to be". 

```suggestion
implementers of new `BioSequence`s might not be optimised.
```

kescobo: Actually, "might not" actually _can_ also mean "not able to", but it is rarely used that way in practice. English is really weird.

docs/src/counting.md
```diff
@@ -6,66 +6,73 @@ end
...
-```jldoctest
-julia> count(!=, dna"ATCGM", dna"GCCGM")
-2
+## Matches and mismatches
+The methods `matches` and `mismatches` take two
+sequences and count the number of positions where the sequences are unequal or equal, respectively.
 
-julia> count(==, dna"ATCGM", dna"GCCGM")
-3
+They are equivalent to `matches(a, b) = count(splat(==), zip(a, b))`
```
kescobo, line 49: I don't know that I've ever seen `splat()` used as a function like that... didn't even know it existed outside of `...` syntax (though now I think about it, it would obviously have to be).

docs/src/counting.md
```diff
@@ -6,66 +6,73 @@ end
...
 
-## Bit-parallel optimisations
+```@docs
+gc_content
+```
 
-For the vast majority of `Base.count(f, seq)` and `Base.count(f, seqa, seqb)`
-methods, a naive counting is done: the internal `count_naive` function is called,
-which simply loops over each position, applies `f`, and accumulates the result.
+## Deprecated aliases
```
kescobo, line 64: I agree with deprecating these in favor of the count methods

kescobo (approved):
Yeah, this all looks great to me. 

The `n_xxxx` functions always seemed weird to me - the `count` methods are much clearer IMO, and documentation is clear about their special-case nature. All :+1: from me!

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "BioSequences"
 uuid = "7e6ae17a-c86d-528c-b3b9-7f778a29fe59"
 authors = ["Sabrina Jaye Ward <sabrinajward@protonmail.com>", "Jakob Nissen <jakobnybonissen@gmail.com>"]
-version = "3.2.0"
+version = "3.3.0"
```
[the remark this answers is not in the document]
jakobnissen: I'll tag 3.4 in an upcoming PR

docs/src/counting.md
jakobnissen: It's new in 1.9 - quite nice for these occasions.

docs/src/counting.md
jakobnissen: This always trips me up in English - despite being fluent, I'm not at the point where these things automatically sound right or wrong :sweat_smile:

PR #325, opened 2024-10-25
Speed up search operations for SeqOrView

jakobnissen:
This commit adds new methods for findnext and findprev for SeqOrView with known alphabets, which use bitparallel operations. This in turns speeds up most search ops which are defined in terms of these.
The new code is 4-20 times faster depending on circumstances.

It's only implemented for known alphabets because new alphabets may overload == in surprising ways, which makes the bitparallel ops invalid.

The commit also introduces a new internal abstraction, the `parts` function, which may be useful for other operations down the line. It's similar to the existing chunk iterators, but may be more efficient for subsequences, and can be reversed.

There is also some minor cleanup that could have been its own PR, but whatever.

## TODO
- [x] Tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@ docs/site/
 Manifest.toml
 
 .DS_Store
+
+LocalPreferences.toml
+
+TODO.md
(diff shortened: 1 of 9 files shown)


jakobnissen:
Codecov is complaining about coverage mostly because parts for LongSequence is not used yet. I have an upcoming PR up that will use it.

PR #332, opened 2025-06-04
Improve GC counting code

jakobnissen:
When implementing https://github.com/BioJulia/Kmers.jl/issues/48, I want to use BioSequences internals. Keep an open branch here for all optimisations I want to upstream.


kescobo:
Any of this that can be shunted off to BioGenerics.jl to avoid the need to duplicate? Or do you mean that BioSequences is a dependency of Kmers, and you want to make the change here instead of there?

PR #338, opened 2025-07-16
Add `spliceinto!` function.

### LINKED ISSUE #336: method for insert!() to add multiple nucleotides
Is it possible to extend the `Base.insert!()` method to allow the insertion of more than a single basepair?

example
```julia
julia> seq = LongDNA{4}("ATACAGGAGAGA")
12nt DNA Sequence:
ATACAGGAGAGA

# works, current method
julia> insert!(seq, 2, 'G')
13nt DNA Sequence:
AGTACAGGAGAGA

# doesn't work, requested method
julia> insert!(seq, 2, dna"GGGG")
ERROR: MethodError: Cannot `convert` an object of type LongSequence{DNAAlphabet{4}} to an object of type DNA
```

My current solution is to loop it, but that seems inefficient, given the copyto! happening under the hood, I think?
```julia
function mutate!(seq::LongDNA{4}, position::Int, replacement::String)
    for (pos,i) in enumerate(replacement)
        insert!(seq, position + pos-1, i)
    end
    return seq
end

julia> seq = LongDNA{4}("ATACAGGAGAGA")
12nt DNA Sequence:
ATACAGGAGAGA

julia> mutate!(seq, 2, "GGG")
15nt DNA Sequence:
AGGGTACAGGAGAGA

```

jakobnissen:
Fixes #336 

@pdimens - I'm interested in feedback. Does this cover your use case, and what do you think of the API?
The reason for the deviation from `splice!` is:
* It bugs me that `4:3 == 5:4`, and yet passing these equal arrays into `splice!` causes different behaviour
* It's unnecessary and inefficient that `splice!` returns the removed sequence

By the way, if you need this function now and can't wait for the 3.5.0 release (which will happen once this PR is merged), or need compat with older versions of BioSequences, you can use the implmentation from this PR. It's efficient, and uses no internals.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [3.5.0]
+* Add `spliceinto!`. This function inserts a sequence into a biosequence,
+  and optionally deletes part of the original sequence. The naming difference
+  from `Base.splice!` reflects is slightly different API.
+* Optimise various methods
+
 ## [3.4.0]
 * Deprecate functions `n_ambiguous`, `n_gaps` and `n_certain`. Instead, use the
   equivalent methods `count(f, seq)` with the appropriate function `f`.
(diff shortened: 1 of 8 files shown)


pdimens:
Thanks, I'll check this in <3hrs. I think having the insertion be a range is redundant, no? Can you see a use case of inserting a collection not at `startposition:start+len`?

pdimens:
Nvm, I read the commit (currently on my phone) and see the optional deletion of destination

jakobnissen:
Yeah, that's also a cool name! Maybe even better.

pdimens:
If it's an indel, it would have to allow the replacement to be of a different length than the original target (be it a single position or a sequential range)

pdimens:
Thank you!

PR #346, opened 2026-01-08
Generalize bit manipulation to larger integers than UInt128

jakobnissen:
This is useful for other packages that may want to hook into this machinery, e.g. Kmers


jakobnissen:
Coverage is fine, and BED failure is unrelated

